PR #53, opened 2023-03-08
Hazard pydantic gw2

GundulaW:
I actually only implemented pydantic in Event and Synthetic, not in Hazard


flood_adapt/object_model/io/unitfulvalue.py
```diff
@@ -0,0 +1,93 @@
...
+        elif self.units == "inch":
+            conversion = 25.4  # millimeters
+        elif self.units == "knots":
+            conversion = 1.0 / 1.943844  # m/s
+        elif self.units == "cfs":  # cubic feet per second
+            conversion = 0.02832  # m3/s
+        else:
+            conversion = 1
+        self.value = self.value * conversion
+        return self
```
Hofer-Julian, line 73: That should be a validator

Hofer-Julian (approved):
Looks good

PR #58, opened 2023-03-09
Change the projection class based on the pydentic and interface classes

panosatha:
Now the risk drivers are not different classes


GundulaW (approved):
all good :)

PR #69, opened 2023-03-14
Database and api pa

panosatha:
- All the pydantic classes are now included in the interface files (solves issues with cylic imports and shows everythings in one place)
- Initial database controller with general methods and measures specific methods.
- Initial API methods to be used in the GUI


flood_adapt/object_model/projection.py
```diff
@@ -53,6 +43,6 @@ def load_dict(data: dict[str, Any]):
         return obj
 
     def save(self, filepath: Union[str, os.PathLike]):
-        """save Projection to a toml file"""
+        """save Elavate to a toml file"""
```
GundulaW, line 46:
```suggestion
        """save projection to a toml file"""
```

tests/test_database.py
GundulaW, line 1: Are you adding some assert lines to the test?

GundulaW (approved):
All good, as we had discussed this beforehand already, so I approve already. 

Only a few minor things:
I replaced some "pass" with "raise NotImplementedError". I added these all as suggestions on GitHub so you can easily accept.

Are you adding some assert statements in the database test?

tests/test_database.py
panosatha: added a small test. Most of the tests are in the API

PR #82, opened 2023-03-26
Create SFINCS model for synthetic event with tide, surge and discharge

GundulaW:
I added discharge boundary consitions to the synthetic event and also moved the test SFINCS model template to the p-drive as it is faster than the n-drive. Perhaps we should move the whole test database there (all toml files) and use the database path that Panos has added to several classes


flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -19,33 +24,70 @@ def load_overland_sfincs_model(self, model_root: Optional[str] = None):
...
         gdf_locs = self.sf_model.forcing["bzs"].vector.to_gdf()
         gdf_locs.crs = self.sf_model.crs
 
         # Go from 1 timeseries to timeseries for all boundary points
-        df_ts = self.tide_ts
-        wl = df_ts[0]
-        for i in range(len(gdf_locs) - 1):
-            col_name = i + 1
-            df_ts[col_name] = wl
+        for i in range(1, len(gdf_locs)):
```
tim-d-w, line 52: This seems incorrect

flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -19,33 +24,70 @@ def load_overland_sfincs_model(self, model_root: Optional[str] = None):
...
+        )
+
+    def write_sfincs_model(self, path_out: Path):
+        """Write all the files for the sfincs model
+
+        Args:
+            path_out (Path): new root of sfincs model
+        """
+        # Change model root to new folder
+        self.sf_model.set_root(path_out, mode="w+")
```
tim-d-w, line 83: set_root does not have a mode

tim-d-w (changes requested):
- ruff and spelling check are not correct according to github (black is correct)

flood_adapt/object_model/hazard/hazard.py
```diff
@@ -102,6 +115,66 @@ def get_event_object(event_path):  # TODO This could be used above as well?
...
+
+        # Run new model (create batch file and run it)
+        # create batch file to run SFINCS, adjust relative path to SFINCS executable for ensemble run (additional folder depth)
+        if self.event.attrs.mode == "risk":
+            with open(run_folder_overland.joinpath("run.bat"), "w") as f_out:
+                bat_file: str = (
+                    "cd "
+                    "%~dp0"
+                    "\n"
+                    f"..\..\..\..\..\..\..\system\sfincs\{site.attrs.sfincs.version}\sfincs.exe"
```
tim-d-w, line 166: python does not like a \{ combination

tim-d-w (approved):

PR #86, opened 2023-03-29
Events tdw

tim-d-w:
made the required functions to add (synthetic) event. Most of my functions are specific for synthetic event, since the pydantic model ISynthetic is different from IEvent.

### DIFF
--- a/flood_adapt/api/events.py
+++ b/flood_adapt/api/events.py
@@ -3,7 +3,8 @@
 from typing import Any
 
 from flood_adapt.dbs_controller import IDatabase
-from flood_adapt.object_model.interface.events import IEvent
+from flood_adapt.object_model.hazard.event.synthetic import Synthetic
+from flood_adapt.object_model.interface.events import IEvent, ISynthetic
 
 
 def get_events(database: IDatabase) -> dict[str, Any]:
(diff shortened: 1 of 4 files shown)


flood_adapt/api/events.py
```diff
@@ -3,13 +3,41 @@
...
+def create_synthetic_event(attrs: dict[str, Any], database: IDatabase) -> ISynthetic:
+    return Synthetic.load_dict(attrs)
+
+
+def save_synthetic_event(event: ISynthetic, database: IDatabase) -> None:
+    database.save_synthetic_event(event)
+
+
+def edit_synthetic_event(event: ISynthetic, database: IDatabase) -> None:
+    database.edit_synthetic_event(event)
```
GundulaW, line 28: I was comparing this against Panos measures code. He does not have a save and edit function for each measure subclass. Can these also be generic save and edit functions for events?

flood_adapt/object_model/interface/events.py
```diff
@@ -75,7 +75,9 @@ class WindModel(BaseModel):
 class RainfallModel(BaseModel):
     source: RainfallSource
     # constant
-    constant_intensity: Optional[UnitfulLength]
+    constant_intensity: Optional[
+        float
+    ]  # TODO: add units; intensity is in mm/hr or in/hr
```
GundulaW, line 80: OK, we leave this as is for now. And I will add this at some point

flood_adapt/dbs_controller.py
```diff
@@ -135,6 +156,33 @@ def save_measure(self, measure: IMeasure) -> None:
                 / f"{measure.attrs.name}.toml"
             )
 
+    def save_synthetic_event(self, event: ISynthetic) -> None:
```
GundulaW, line 159: Can this be generic for all event types?

flood_adapt/dbs_controller.py
```diff
@@ -151,6 +199,19 @@ def edit_measure(self, measure: IMeasure):
             / f"{measure.attrs.name}.toml"
         )
 
+    def edit_synthetic_event(self, event: ISynthetic):
```
GundulaW, line 202: can this be the same for all event types?

GundulaW (approved):
I've approved with a few questions: Can the save and edit functions be the same for all event types? 

And I am missing a test, there is a separate folder for api tests where you can add that. You can check Panos code or that

PR #91, opened 2023-03-31
Latest version with most of api finished

panosatha:


flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -21,11 +22,11 @@ def __init__(self, model_root: Optional[str] = None):
         # if validate_existence_root_folder(model_root):
         #    self.model_root = model_root
 
-        self.sf_model = SfincsModel(root=model_root, mode="r+")
+        # self.sf_model = SfincsModel(root=model_root, mode="r+")
```
GundulaW, line 25: Do we have a fix for the hydromt issue yet?

We only need to comment out the import of the SfincsModel from hydromt. The rest can remain in the script. At least this worked for me

flood_adapt/object_model/interface/database.py
GundulaW, line 1: projection and event methods (get/save/edit/delete/copy) are still missing

GundulaW (approved):
1. Do we have a fix for the hydromt issue yet?
We only need to comment out the import of the SfincsModel from hydromt. The rest can remain in the script. At least this worked for me

2. The get/save/edit/delete/copy methods are missing in the database interface class, should be quick to fix

flood_adapt/integrator/sfincs_adapter.py
panosatha: we install the correct hydromt version with the environment yml file to make sure that it works.

PR #99, opened 2023-04-10
Sfincs button gw

GundulaW:
If you want the sfincs tests to work in pytest, you need to place the SFINCS template and executable folder locally in the FloodAdapt folder. I also updated the site.toml, committed that and fixed some tests. But I do not have the test_database-GUI folder, so I could not run the api tests


tim-d-w (commented):
I had to change the file path of the sfincs executable which is made in the run.bat file or else the path was wrong (I had to delete this: '..\'), so I think we went one directory up too far. If this was working for you Gundula, let me know!

GundulaW:
Hi Tim, 

For me this worked without changing the path in the batch file. Is your SFINCS executable in the right folder? There is a separate folder in the system\sfincs folder for each SFINCS release, see below:
This is where it should be \FloodAdapt-GUI\system\sfincs\sfincs20_AlpeDHuez_release\sfincs.exe

PR #107, opened 2023-05-08
Fiat adapter changes

panosatha:


Mares2022 (approved):
Fiat adaptor including buyout measures and first test of set_hazard

PR #114, opened 2023-05-26
merged demo with main and all test passes

srpeiter:


environment.yml
```diff
@@ -2,12 +2,18 @@ name: flood_adapt
 
 channels:
   - conda-forge
+  - numba
```
Hofer-Julian, line 5: Why do we need a numba channel?

Hofer-Julian (commented):
I am a bit confused by the numba channel, otherwise it looks good

environment.yml
srpeiter: Reason is that noaa-coops uses numpy >= 1.24 which is supported by numba 0.57. The latest version of numba i.e. 0.57 is not yet available on conda-forge, but on numba

Hofer-Julian (approved):

PR #115, opened 2023-06-01
small correction in an assertion error in the measures test

panosatha:


srpeiter:
First teamcity built failed, but after re running the built, it went ok

PR #134, opened 2023-06-27
Risk mode gw

GundulaW:


flood_adapt/object_model/hazard/hazard.py
```diff
@@ -128,78 +177,260 @@ def add_discharge(self):
...
+        # Preprocess all hazard model input
+        self.preprocess_sfincs()
+        # add other models here
+
+    def postprocess_models(self):
+        # Preprocess all hazard model input
+        self.postprocess_sfincs()
+        # add other models here
 
     def run_models(self):
```
panosatha, line 198: shouldn't there be a seperate run_sfincs method that is run here? The same way you do for preprocess for example

flood_adapt/object_model/scenario.py
```diff
@@ -57,13 +57,34 @@ def save(self, filepath: Union[str, os.PathLike]):
...
+            self.direct_impacts.hazard.run_models()
         if not self.direct_impacts.has_run:
             self.direct_impacts.run_models()
         else:
             print(
                 f"Direct impacts for scenario '{self.attrs.name}' has already been run."
             )
+
+        if not self.direct_impacts.hazard.has_run:
+            self.direct_impacts.hazard.postprocess_models()
```
panosatha, line 83: isn't this needed before you run direct impacts? At least for the risk run where the RP flood maps are needed as input

panosatha:
check ruff, typos and TeamCity tests

panosatha (changes requested):
Good job! I added some comments that you can have a look before merging

flood_adapt/object_model/scenario.py
GundulaW: I included all functions in one if statement: 

        if not self.direct_impacts.hazard.has_run:
            self.direct_impacts.hazard.preprocess_models()
            self.direct_impacts.hazard.run_models()
            self.direct_impacts.hazard.postprocess_models()
        else:
            print(f"Hazard for scenario '{self.attrs.name}' has already been run.")

PR #140, opened 2023-06-28
Historical offshore tdw

tim-d-w:


GundulaW (approved):
do we need/want to use interface classes for the individual event types?

PR #163, opened 2023-07-07
Feature/split hurricane backend frontend

tim-d-w:


flood_adapt/object_model/hazard/hazard.py
```diff
@@ -207,14 +209,17 @@ def get_event_object(event_path):  # TODO This could be used above as well?
             return EventSet.load_file(event_path)
 
     def preprocess_models(self):
+        print("Preparing hazard models...")
```
dladrichem, line 212: If these are left here on purpose, please use the logging library:
import logging
logging.debug/info/warning/error

dladrichem: Same for all other print statements

dladrichem (approved):

PR #165, opened 2023-07-07
131 urban green infra backend

tim-d-w:


flood_adapt/object_model/hazard/measure/green_infrastructure.py
```diff
@@ -0,0 +1,106 @@
...
+        -------
+        float
+
+
+        Returns
+        -------
+        float
+            Volume [m3]
+        """
+        volume = area * height * percent_area / 100.0
```
GundulaW, line 79: Why is there no unit conversion applied? Where is this function used? 

```suggestion
        volume = area.convert("m2") * height.convert("meters") * percent_area / 100.0
```

flood_adapt/object_model/hazard/hazard.py
```diff
@@ -317,16 +317,20 @@ def preprocess_sfincs(
...
+                measure_path = base_path.joinpath(
+                    "input", "measures", measure.attrs.name
+                )
+                if measure.attrs.type == "floodwall":
+                    model.add_floodwall(
+                        floodwall=measure.attrs, measure_path=measure_path
+                    )
+                if measure.attrs.type == "green_infrastructure":
+                    model.add_green_infrastructure(
+                        green_infrastructure=measure.attrs, measure_path=measure_path
```
GundulaW, line 332: This is missing one indent and will only write out the last SFINCS model for an event set
```suggestion
          # Add floodwall if included
          if self.hazard_strategy.measures is not None:
              for measure in self.hazard_strategy.measures:
                  measure_path = base_path.joinpath(
                      "input", "measures", measure.attrs.name
                  )
                  if measure.attrs.type == "floodwall":
                      model.add_floodwall(
                          floodwall=measure.attrs, measure_path=measure_path
                      )
                  if measure.attrs.type == "green_infrastructure":
                      model.add_green_infrastructure(
                          green_infrastructure=measure.attrs, measure_path=measure_path
```

tests/test_object_model/test_measures.py
```diff
@@ -184,3 +188,45 @@ def test_floodproof_read():
...
+
+    # def test_calculate_area():
+    #     test_toml = (
+    #         test_database
+    #         / "charleston"
+    #         / "input"
+    #         / "measures"
+    #         / "green_infra"
+    #         / "green_infra.toml"
+    #     )
```
GundulaW, line 232: add more meaningful tests

GundulaW (changes requested):
All comments include suggestions so you can easily accept them in github. There could be more meaningful tests but you can skip this for now and I will create an issue.

There is an indent missing in hazard.py. That one is important because for an event set it will only write out the last SFINCS model at the moment.

The other ones are minor things in function descriptions.

GundulaW (approved):
Good job!

PR #176, opened 2023-07-12
145 infographic completion backend

### LINKED ISSUE #145: Infographic completion backend
### Kind of request

Adding new functionality

### Enhancement Description

In order for the infographic to run, the backend needs to be fixed first

### Task list

- [x] Implement fiat_toolbox metrics writer
- [x] Create infographic script
- [x] Unit test
- [x] Test with GUI

### Use case

_No response_

### Additional Context

_No response_

dladrichem:
Fixes #145


flood_adapt/api/output.py
```diff
@@ -1,6 +1,8 @@
+from pathlib import Path
 from typing import Any
 
 import pandas as pd
```
srpeiter, line 4: Please run "ruff check --fix ." in the project folder.
This will show you all these type of errors and will try to fix it for you.

flood_adapt/api/output.py
```diff
@@ -36,8 +38,18 @@ def get_aggregation(name: str, database: IDatabase):
 
 
 def make_infographic(name: str, database: IDatabase) -> str:
-    return database.get_scenario(name).infographic()
+    # Get the scenario
+    scenario = database.get_scenario(name)
+
+    # Check if the scenario has run
+    scenario.has_run_impact = (
```
srpeiter, line 45: This is not checking, but an assignment?

flood_adapt/api/output.py
srpeiter: Suggestion: maybe we should try implementing pre-commit hooks, such that these standard actions are triggered automatically when committing.

srpeiter, line 4: Please run "ruff check --fix ." in the project folder.
This will show you all the linting errors and will try to fix it for you.
 and "black ." to do automatic formatting.

Suggestion: maybe we should try implementing pre-commit hooks, such that these standard actions are triggered automatically when committing.

flood_adapt/api/output.py
srpeiter, line 45: This is not checking, but an assignment?
Is it what you intended to do?

dladrichem: It should later be a check but from what I understood, this isn't implemented yet so that is why it is now an assignment. Is this implemented now in the FIAT adapter?

srpeiter: Ohw ok, I see.
I think it already happens in floodadapt: https://github.com/Deltares/FloodAdapt/blob/main/flood_adapt/object_model/direct_impacts.py#L89C9-L89C19

dladrichem: Als het goed is zo gefixt

srpeiter (approved):

PR #184, opened 2023-07-17
142 add rainfall and river discharge plots

GundulaW:


flood_adapt/api/__init__.py
panosatha, line 1: why is this needed?

.gitignore
```diff
@@ -171,6 +171,8 @@ tests/test_database/charleston/static/templates/overland/
 tests/test_database/charleston/input/measures/raise_property_aggregation_area/raise_property_aggregation_area_new.toml
 tests/system/sfincs/
 database
+tests/test_database/charleston/static/cyclone_track_database/
+tests/test_object_model/test_events.py
 tests/test_database/charleston/input/events/kingTideNov2021/gfs_anl_0p50_us_southeast.20211104_0000.nc
```
panosatha, line 176: can't you use wildcards instead of having each file? something like *.nc

flood_adapt/integrator/__init__.py
panosatha, line 1: do we still need this file?

flood_adapt/object_model/hazard/event/event.py
```diff
@@ -213,38 +180,52 @@ def add_dis_ts(self):
             updated object with wind timeseries added in pf.DataFrame format
```
panosatha, line 180: replace wind with discharge?

flood_adapt/object_model/hazard/hazard.py
```diff
@@ -29,6 +31,7 @@ class Hazard:
     database_input_path: Path
     mode: Mode
     event_set: EventSet
```
panosatha, line 33: This is double

```suggestion
```

flood_adapt/object_model/interface/site.py
```diff
@@ -46,6 +46,12 @@ class SfincsModel(BaseModel):
     floodmap_units: UnitTypesLength
 
 
+class Cyclone_track_databaseModel(BaseModel):
```
panosatha, line 49: do we need a model for a single attribute of a string?

panosatha:
@GundulaW, test_constant_rainfal in test_events.py fails for me. The reason seems to be that you are asserting the value of intensity in inches while in the backend you transform it to mm.

panosatha (changes requested):
Almost there except the test that is failing. I did not go through the cyclone related changes, since they are the focus of the other pull request.

.gitignore
GundulaW: done, but my test database seems screwed up. None of the FIAT or benefits tests are working for me.

flood_adapt/object_model/hazard/hazard.py
GundulaW: done

PR #187, opened 2023-07-17
129 add delete protection for events projections scenarios and strategies

dladrichem:
This fixes both the delete protection and the standard events. It is linked to GUI issue 78

### DIFF
--- a/flood_adapt/dbs_controller.py
+++ b/flood_adapt/dbs_controller.py
@@ -475,8 +475,8 @@ class Database(IDatabase):
         """
 
         # Check if event is a standard event
-        if self.site.attrs.standard_events:
-            if name in self.site.attrs.standard_events:
+        if self.site.attrs.standard_objects.events:
+            if name in self.site.attrs.standard_objects.events:
                 raise ValueError(
                     f"'{name}' event cannot be deleted since it is a standard event."
                 )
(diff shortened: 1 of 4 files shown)


flood_adapt/dbs_controller.py
```diff
@@ -748,16 +813,35 @@ def delete_scenario(self, name: str):
...
             )
         else:
-            shutil.rmtree(scenario_path, ignore_errors=True)
+            scenario_path = self.input_path / "scenarios" / name
+            scenario = Scenario.load_file(scenario_path / f"{name}.toml")
+            scenario.init_object_model()
+            if scenario.direct_impacts.hazard.has_run:
+                # TODO this should be a check were if the scenario is run you get a warning?
+                raise ValueError(
+                    f"'{name}' scenario cannot be deleted since the hazard model has already run."
```
panosatha, line 841: maybe better return "the scenario has been run" instead of "the hazard model has been run"

flood_adapt/object_model/interface/site.py
```diff
@@ -153,6 +153,9 @@ class SiteModel(BaseModel):
     obs_station: Optional[Obs_stationModel]
     benefits: BenefitsModel
     scs: Optional[SCSModel]  # optional for the US to use SCS rainfall curves
+    standard_events: Optional[list[str]]
+    standard_projections: Optional[list[str]]
+    standard_strategies: Optional[list[str]]
```
panosatha, line 158: maybe better to have a model named StandardObjectsModel where you have these three lists?

panosatha (commented):
nice work! I made some minor comments, which you can check before merging

PR #188, opened 2023-07-17
186 wrong slr calculations

GundulaW:
Calculation was right (including the SLR up until the reference year). But in the plot I did not subtract the SLR up until the reference year, hence the interpolated value and plot were inconsistent. I fixed the plot now to make it consistent. I had discussed this with Kathryn

### DIFF
--- a/flood_adapt/dbs_controller.py
+++ b/flood_adapt/dbs_controller.py
@@ -89,6 +89,27 @@ class Database(IDatabase):
         return df.columns[2:].to_list()
 
     def interp_slr(self, slr_scenario: str, year: float) -> float:
+        """interpolating SLR value and referencing it to the SLR reference year from the site toml
+
+        Parameters
+        ----------
+        slr_scenario : str
+            SLR scenario name from the coulmn names in static\slr\slr.csv
+        year : float
+            year to evaluate
+
+        Returns
+        -------
+        float
+            _description_
+
+        Raises
+        ------
+        ValueError
+            if the reference year is outside of the time range in the slr.csv file
+        ValueError
+            if the year to evaluate is outside of the time range in the slr.csv file
+        """
         input_file = self.input_path.parent.joinpath("static", "slr", "slr.csv")
         df = pd.read_csv(input_file)
         if year > df["year"].max() or year < df["year"].min():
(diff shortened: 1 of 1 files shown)


flood_adapt/dbs_controller.py
```diff
@@ -133,13 +154,23 @@ def plot_slr_scenarios(self) -> str:
...
+        else:
+            scenarios = self.get_slr_scn_names()
+            for scn in scenarios:
+                ref_slr = np.interp(ref_year, df["Year"], df[scn])
+                df[scn] -= ref_slr
+
         df = df.drop(columns="units").melt(id_vars=["Year"]).reset_index(drop=True)
         # convert to units used in GUI
-        slr_current_units = UnitfulLength(value=df.iloc[0, -1], units=units)
+        slr_current_units = UnitfulLength(value=1.0, units=units)
```
dladrichem, line 170: Wouldn't it be prettier if we moved most of the unit conversion to the type class (the conversion factor has nothing to do with the value so calculating the conversion factor can be moved to in this case UnitTypesLength). In that case, you also don't need the silly unit length declaration to get the conversion factor

flood_adapt/dbs_controller.py
```diff
@@ -133,13 +154,23 @@ def plot_slr_scenarios(self) -> str:
...
         gui_units = self.site.attrs.gui.default_length_units
-        slr_gui_units = slr_current_units.convert(gui_units)
-        conversion_factor = slr_gui_units / slr_current_units.value
-        df.iloc[:, -1] = conversion_factor * df.iloc[:, -1]
+        conversion_factor = slr_current_units.convert(gui_units)
+        df.iloc[:, -1] = (conversion_factor * df.iloc[:, -1]).round(decimals=2)
```
dladrichem, line 173: Also you can do this all with a single line. It should be something like:
 df['value'] =df.apply(lambda row : round(UnitfulLength(value = row['value'], units = units).value, ndigits = 2), axis = 1)

flood_adapt/dbs_controller.py
```diff
@@ -204,15 +236,10 @@ def plot_wl(self, event: IEvent, input_wl_df: pd.DataFrame = None) -> str:
...
-                value=float(wl_df.iloc[0, 0]), units="meters"
-            )
+            wl_current_units = UnitfulLength(value=1.0, units="meters")
             gui_units = self.site.attrs.gui.default_length_units
-            wl_gui_units = wl_current_units.convert(gui_units)
-            if wl_current_units.value == 0:
-                conversion_factor = 1
-            else:
-                conversion_factor = wl_gui_units / wl_current_units.value
+            conversion_factor = wl_current_units.convert(gui_units)
```
dladrichem, line 241: Same kind of lambda function can be used here

dladrichem (changes requested):

flood_adapt/dbs_controller.py
dladrichem: Check next comment first, moving it might still be prettier but that is only for the lower level class

flood_adapt/dbs_controller.py
dladrichem: df.apply and lambda functions allow you to run a function on every row separately, which is what you're trying to do here I assume

flood_adapt/dbs_controller.py
GundulaW: I'll make an issue out of it. Kathryn needs to have this fixed for the demo

dladrichem (approved):

PR #193, opened 2023-07-28
aggregation areas and measures

panosatha:
- Updated database.get_aggregation_areas method to return a dictionary of the different aggregation areas (this is more usefull for getting the correct type of aggregation areas)
- Added a functionality in database.get_measures method to return a geometry for the impact measures based on an aggregation area

### DIFF
--- a/flood_adapt/dbs_controller.py
+++ b/flood_adapt/dbs_controller.py
@@ -59,7 +59,7 @@ class Database(IDatabase):
         # self.update()
 
     # General methods
-    def get_aggregation_areas(self) -> list[GeoDataFrame]:
+    def get_aggregation_areas(self) -> dict:
         """Get a list of the aggregation areas that are provided in the site configuration.
         These are expected to much the ones in the FIAT model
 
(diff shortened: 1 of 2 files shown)


flood_adapt/dbs_controller.py
```diff
@@ -968,12 +969,27 @@ def get_measures(self) -> dict[str, Any]:
...
+
+        geometries = []
+        for path, obj in zip(measures["path"], objects):
+            # If polygon is used read the polygon file
+            if obj.attrs.polygon_file is not None:
+                geometries.append(
+                    gpd.read_file(path.parent.joinpath(obj.attrs.polygon_file))
+                )
+            # If aggregation area is used read the polygon from the aggregation area name
+            elif obj.attrs.aggregation_area_name is not None:
```
dladrichem, line 981: Same as in the GUI review, is not none is not necessary

flood_adapt/dbs_controller.py
```diff
@@ -968,12 +969,27 @@ def get_measures(self) -> dict[str, Any]:
...
+                # TODO Maybe it makes more sense to have the geometry returned here?
+                aggr_areas = self.get_aggregation_areas()
```
dladrichem, line 983: Running this once before the loop would already help a lot I assume in speeding up the code. You don't need to call it every loop. get_aggregation_areas() is a very slow function since it reads the entire file and converts all aggregation areas to another crs every time you run

flood_adapt/dbs_controller.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, database_path: Union[str, os.PathLike], site_name: str) -> No
         # self.update()
 
     # General methods
-    def get_aggregation_areas(self) -> list[GeoDataFrame]:
+    def get_aggregation_areas(self) -> dict:
```
dladrichem, line 62: Wouldn't it be useful to also have a function where you can get a single area instead of doing all transitions on all area's

dladrichem (changes requested):

dladrichem (approved):

PR #194, opened 2023-08-03
Output metrics and comparison tab backend

dladrichem:


flood_adapt/object_model/scenario.py
```diff
@@ -73,6 +76,55 @@ def run(self):
                 f"Direct impacts for scenario '{self.attrs.name}' has already been run."
             )
 
+        # Create the infometrics and infographic files
+        self._create_infometrics_infographics()
+
+    def _create_infometrics_infographics(self):
+        # Get the metrics
+        metrics_results_path = self.database_input_path.parent.joinpath(
```
panosatha, line 84: this is the fiat results path. Why is it named metrics_results_path?

flood_adapt/object_model/scenario.py
```diff
@@ -73,6 +76,55 @@ def run(self):
                 f"Direct impacts for scenario '{self.attrs.name}' has already been run."
             )
 
+        # Create the infometrics and infographic files
+        self._create_infometrics_infographics()
+
+    def _create_infometrics_infographics(self):
```
panosatha, line 82: wouldn't it make more sense to have two seperate methods for the creation of the metrics table and the creatio of the infographic?

flood_adapt/object_model/scenario.py
```diff
@@ -73,6 +76,55 @@ def run(self):
...
+
+        metrics_writer.parse_metrics_to_file(
+            df_results=df, metrics_path=metrics_outputs_path, write_aggregate=None
+        )
+        metrics_writer.parse_metrics_to_file(
+            df_results=df, metrics_path=metrics_outputs_path, write_aggregate="all"
+        )
+
+        # Get the infographic
+        InfographicsParser().write_infographics_to_file(
```
panosatha, line 122: It seems that while the InfographicsParser class is in fiat_toolbox, it is really tailoed to the floodadapt database structure. Wouldn't it make more sense to not rely on this structure since this can be used outside of floodadapt as well? 

I would expect that the infographic method would need the metrics files, and a configuration file only. The database paths should only be defined here in this floodadapt method.

flood_adapt/object_model/scenario.py
```diff
@@ -73,6 +76,55 @@ def run(self):
...
+            scenario_name=self.attrs.name,
+            database_path=Path(self.database_input_path).parent,
+            keep_metrics_file=True,
+        )
```
panosatha, line 126: Additionally, I see that the infographic creation files are now in the input folder, while I think it would make more sense for them to be in the static folder, since they are not part of the general floodadapt object model.

panosatha (changes requested):
Nice work! I thing some parts mainly related to the folder structure need to be changed. See my comments.

I tried to make an infometric with the total damage (using SUM) and the infographic creation broke, since with the new float value, its was reading all the values as strings. I think the type of each value might be needed as input in the config.

Additionally, I didn't see any unit test for the infometrics and infographics. 

Task list:
- [ ] Change location of infographic config (from input to static)
- [ ] Change output folders and naming (do we need the seperate infographic and infometric folders?)
- [x] Seperate infometric and infographic creations in method
- [ ] Make sure that the fiat_toolbox operation are disconnected to the floodadapt folder structure
- [x] Make sure that value type (float or int) is correctly read
- [ ] Add unit tests

PR #195, opened 2023-08-08
98 remove long name in back end

### LINKED ISSUE #98: Remove long_name in back-end

dladrichem:
Closes #98 
Also while chaning the long_name I broke the tests. It was quite difficult to get it back to the original state since a lot of files were gitignored and the tests also create files. Therefore, I added a fixture that deletes all files that are created while running the test


GundulaW (approved):
all good

PR #202, opened 2023-08-28
167 make sure benefit calculation works in the real workflow

panosatha:


tests/test_database/charleston/static/templates/fiat/Exposure/exposure.gpkg
GundulaW, line 1: This seems like a big file. I think we should leave this on the p-drive

GundulaW (approved):
small comments. Just do another pull from env_fix

flood_adapt/object_model/scenario.py
```diff
@@ -80,7 +80,9 @@ def run(self):
         self._create_infometrics()
 
         # Create the infographic files
-        self._create_infographics()
+        # TODO correct infographic creation for risk mode
+        if self.direct_impacts.hazard.event_mode != "risk":
+            self._create_infographics()
```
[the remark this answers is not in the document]
panosatha: indeed but for now I added this so we dont get an error

tests/test_database/charleston/static/templates/fiat/Exposure/exposure.gpkg
panosatha: good catch. I added it by accident

PR #215, opened 2023-08-31
fixing unit conversion for timeseries

GundulaW:
- [x] closing https://github.com/Deltares/FloodAdapt/issues/205
- [x] closing https://github.com/Deltares/FloodAdapt/issues/126


GundulaW:
I decided to not save time series files in metric units but instead keep everything in the backend in gui units and only convert units in the sfincs adapter

PR #234, opened 2023-09-08
196 create a log file for scenario simulation progress

GundulaW:
Hey Daley,

Is this how you would do this? And would this be interfering with your GUI logging?


flood_adapt/object_model/scenario.py
```diff
@@ -16,7 +17,7 @@ class Scenario(IScenario):
 
     attrs: ScenarioModel
     direct_impacts: DirectImpacts
-    database_input_path: Union[str, os.PathLike]
```
dladrichem, line 19: Why not keep the string option? You can always convert the string to a path object (I personally prefer Pathlib over os.PathLike) first thing in the function

flood_adapt/object_model/scenario.py
```diff
@@ -57,6 +58,23 @@ def save(self, filepath: Union[str, os.PathLike]):
...
+        )
+        for parent in reversed(results_dir.parents):
+            if not parent.exists():
+                os.mkdir(parent)
+        if not results_dir.exists():
+            os.mkdir(results_dir)
+        # Initiate the logger for all the integrator scripts.
+        self.initiate_root_logger(results_dir.joinpath(f"{self.attrs.name}.log"))
+        version = "0.1.0"
+        logging.info(f"FloodAdapt version {version}")
```
dladrichem, line 73: It shouldn't interfere since you are not including the overall logger (which is also not possible since this is the backend)

flood_adapt/object_model/scenario.py
```diff
@@ -82,3 +100,28 @@ def __eq__(self, other):
         test2 = self.attrs.projection == other.attrs.projection
         test3 = self.attrs.strategy == other.attrs.strategy
         return test1 & test2 & test3
+
+    @staticmethod
+    def initiate_root_logger(filename):
```
dladrichem, line 105: In the future, it would be nice to create a similar backend logger to capture stuff like this. But for now, this is fine

flood_adapt/object_model/scenario.py
GundulaW: reverted this change

PR #239, opened 2023-09-14
Multiple fixes for infographics

### LINKED ISSUE #206: Infometrics CSV file is written twice
### FloodAdapt version checks

### Reproducible Example

If you run a scenario, the metrics CSV is written, then the metrics CSV at aggregation scale, then the metrics CSV is written again.

### Current behaviour

If you run a scenario, the metrics CSV is written, then the metrics CSV at aggregation scale, then the metrics CSV is written again.

### Desired behaviour

The metrics CSV should only be generated once

### Task list

* Find out why the metrics CSV is being written a second time
* Correct the problem

### Additional context

_No response_

### LINKED ISSUE #219: move infometric and infographic methods to direct_impacts
### Kind of request

### Enhancement Description

Right now these methos are in the scenario class but wouldn't it make more sense for them to be in the direct_impacts one, as prostprocessing steps of FIAT?

### Task list

TBD

### Use case

_No response_

### Additional Context

_No response_

dladrichem:
Closes #219 & Closes  #206

### DIFF
--- a/flood_adapt/object_model/direct_impacts.py
+++ b/flood_adapt/object_model/direct_impacts.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import geopandas as gpd
 import pandas as pd
 from fiat_toolbox.infographics.infographics import InfographicsParser
+from fiat_toolbox.infographics.infographics_factory import InforgraphicFactory
 from fiat_toolbox.metrics_writer.fiat_write_metrics_file import MetricsFileWriter
 from fiat_toolbox.spatial_output.aggregation_areas import AggregationAreas
 from fiat_toolbox.spatial_output.points_to_footprint import PointsToFootprints
@@ -126,6 +127,8 @@ class DirectImpacts:
         end_time = time.time()
         print(f"Running FIAT took {str(round(end_time - start_time, 2))} seconds")
 
+        self.postprocess_fiat()
+
         # Indicator that direct impacts have run
         if return_code == 0:
             self.__setattr__("has_run", True)
@@ -243,6 +246,7 @@ class DirectImpacts:
             return process.returncode
 
     def postprocess_fiat(self):
+        # Get the metrics
         fiat_results_path = self.database_input_path.parent.joinpath(
             "output",
             "results",
@@ -255,9 +259,7 @@ class DirectImpacts:
         self._create_infometrics(fiat_results_path)
 
         # Create the infographic files
-        # TODO correct infographic creation for risk mode
-        if self.hazard.event_mode != "risk":
-            self._create_infographics()
+        self._create_infographics(self.hazard.event_mode)
 
         # Aggregate results to regions
         self._create_aggregation()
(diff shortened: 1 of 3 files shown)


flood_adapt/object_model/direct_impacts.py
```diff
@@ -126,6 +126,8 @@ def run_models(self):
         end_time = time.time()
         print(f"Running FIAT took {str(round(end_time - start_time, 2))} seconds")
 
+        self.postprocess_fiat()
```
panosatha, line 129: Why was this added here? I think the whole idea was to seperate preproccess, run, and postprocess to different methods. 

The scenario.run() method takes care of the workflow.
```suggestion
```

flood_adapt/object_model/direct_impacts.py
```diff
@@ -243,6 +245,7 @@ def run_fiat(self):
             return process.returncode
 
     def postprocess_fiat(self):
+        # Get the metrics
```
panosatha, line 248:
```suggestion
        # Get FIAT results
```

flood_adapt/object_model/direct_impacts.py
```diff
@@ -373,14 +374,24 @@ def _create_infometrics(self, fiat_results_path):
...
+    def _create_infographics(self, mode):
         # Get the infographic
-        InfographicsParser().write_infographics_to_file(
-            scenario_name=self.name,
-            database_path=Path(self.database_input_path).parent,
-            keep_metrics_file=True,
+        database_path = Path(self.database_input_path).parent
+        metrics_path = database_path.joinpath(
+            "output", "infometrics", f"{self.name}_metrics.csv"
         )
```
panosatha, line 388: couldn't this be saved in the class when _create_infometrics is run? This way we make sure that the path is defined once.

panosatha:
- [ ] It seems that the api that gets the location of the infographic has not been updated accordingly and the infographic is not loaded in the collapsable frame. **api.output.get_infographic()** should be changed
- [ ] I see that there is a **api.output.make_infographic()** as well. Do we need that?
- [ ] Do a merge of env_fix, because I think not everything is up to date

flood_adapt/object_model/direct_impacts.py
dladrichem: You're right, some stuff went wrong with the rebase apparently

panosatha (changes requested):
See my previous comments and in addition:

- [x] It seems that the api that gets the location of the infographic has not been updated accordingly and the infographic is not loaded in the collapsable frame. api.output.get_infographic() should be changed
- [ ]  I see that there is an api.output.make_infographic() as well. Do we need that?
- [x]  Do a merge of env_fix, because I think not everything is up to date

flood_adapt/object_model/direct_impacts.py
```diff
@@ -373,14 +374,24 @@ def _create_infometrics(self, fiat_results_path):
         metrics_writer.parse_metrics_to_file(
             df_results=df, metrics_path=metrics_outputs_path, write_aggregate=None
         )
-        metrics_writer.parse_metrics_to_file(
-            df_results=df, metrics_path=metrics_outputs_path, write_aggregate="all"
-        )
 
-    def _create_infographics(self):
+        if self.hazard.event_mode != "risk":
```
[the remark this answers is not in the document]
dladrichem: Same thing, I didn't need it and didn't think anyone else was using it

PR #246, opened 2023-09-21
245 water level offset

GundulaW:
This is ready for review. 

Only the benefit test failed: 
./tests/test_api/test_benefits.py::test_benefit Failed: [undefined]KeyError: 'TotalEAD'
self = Index(['ExpectedAnnualDamages', 'FloodedHomes', 'ImpactedHomes2Y',
       'ImpactedHomes5Y', 'ImpactedHomes10Y', 'Impa...lDamageRP5', 'TotalDamageRP10', 'TotalDamageRP25',
       'TotalDamageRP50', 'TotalDamageRP100'],
      dtype='object')
key = 'TotalEAD', method = None, tolerance = None

familiar with that?

### DIFF
--- a/flood_adapt/api/output.py
+++ b/flood_adapt/api/output.py
@@ -60,11 +60,13 @@ def get_infographic(name: str, database: IDatabase) -> str:
         raise ValueError(
             f"Scenario {name} has not been run. Please run the scenario first."
         )
-    
+
     database_path = Path(database.input_path).parent
     config_path = database_path.joinpath("static", "templates", "infographics")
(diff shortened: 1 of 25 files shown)


GundulaW:
This is closing https://github.com/Deltares/FloodAdapt/issues/245

GundulaW:
I checked the numbers against Kathryn's kingtide run with the CFRSS and the bzs file is now the same

panosatha:
the issue with the benefits tests origineates from a change in the names in the risk metrics that was not aligned with the benefits calcs. Have fixed that now in my equity branch that I will submit today.

PR #247, opened 2023-09-27
242 allow deleting scenario if it has been run

### LINKED ISSUE #242: Allow deleting scenario if it has been run
### Kind of request

### Enhancement Description

Right now, it is not possible to delete a scenario if it has been run. This should not be the case. A scenario should only have delete protection if is is already used in a benefits analysis

### Task list

- [x] Allow scenario deleting
- [x] Test

### Use case

_No response_

### Additional Context

_No response_

dladrichem:
Closes #242 

Just to clarify, I added the logger cleanup as a destructor to make sure that you only and always clean it up at the end of using the adapter. From what I could see, the single adapter instance is not used anymore after the positions I deleted them right? I did check it but that might be something to check carefully for the reviewer

### DIFF
--- a/flood_adapt/dbs_controller.py
+++ b/flood_adapt/dbs_controller.py
@@ -13,8 +13,8 @@ import xarray as xr
 from cht_cyclones.tropical_cyclone import TropicalCyclone
 from geopandas import GeoDataFrame
 from hydromt_fiat.fiat import FiatModel
-from hydromt_sfincs import SfincsModel
 
+from flood_adapt.integrator.sfincs_adapter import SfincsAdapter
 from flood_adapt.object_model.benefit import Benefit
 from flood_adapt.object_model.hazard.event.event import Event
 from flood_adapt.object_model.hazard.event.event_factory import EventFactory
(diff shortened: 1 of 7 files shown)


flood_adapt/dbs_controller.py
```diff
@@ -1149,15 +1149,11 @@ def delete_scenario(self, name: str):
...
-                # TODO this should be a check were if the scenario is run you get a warning?
-                raise ValueError(
-                    f"'{name}' scenario cannot be deleted since the scenario has been run."
-                )
-            else:
-                shutil.rmtree(scenario_path, ignore_errors=True)
+            shutil.rmtree(scenario_path, ignore_errors=False)
+
+            results_path = self.input_path.parent / "output" / "Scenarios" / name
+            if results_path.exists():
```
panosatha, line 1155: shouldn't there be a warning here that we can use in the front end, in case the scenario has already been run?

flood_adapt/dbs_controller.py
```diff
@@ -1149,15 +1149,11 @@ def delete_scenario(self, name: str):
...
-            scenario = Scenario.load_file(scenario_path / f"{name}.toml")
-            scenario.init_object_model()
-            if scenario.direct_impacts.hazard.has_run:
-                # TODO this should be a check were if the scenario is run you get a warning?
-                raise ValueError(
-                    f"'{name}' scenario cannot be deleted since the scenario has been run."
-                )
-            else:
-                shutil.rmtree(scenario_path, ignore_errors=True)
+            shutil.rmtree(scenario_path, ignore_errors=False)
```
panosatha, line 1152: replace strategy with scenario throughout this method

panosatha (changes requested):
I suggested some small changes in the databse controller, mainly to produce a warning message

flood_adapt/dbs_controller.py
dladrichem: Sharp 👍

PR #268, opened 2023-10-13
267 elevate buildings with datum ref

panosatha:


tests/test_integrator/test_fiat_adapter.py
```diff
@@ -172,7 +172,24 @@ def test_fiat_adapter_measures(cleanup_database):
...
+        / "current_extreme12ft_raise_datum"
+        / "current_extreme12ft_raise_datum.toml"
+    )
+
+    assert test_toml.is_file()
+
+    # use event template to get the associated Event child class
+    test_scenario = Scenario.load_file(test_toml)
+    test_scenario.run()
+
```
GundulaW, line 190: ideally you would do just a unit test and test this one function. E.g. that certain buildings in the exposure csv are higher now after applying the measure. 
Because the model is run it takes really long and we are not checking whether the measure was actually applied.

GundulaW (approved):
I would just change the test to check whether the buildings were elevated by the right amount. Now you are just checking whether FIAT still runs, which takes a lot of time. I would remove the scenario.run() and instead read in the exposure csv to check whether buildings have been elevated by the right amount

PR #285, opened 2023-11-02
200 create geojson file for green infra measure for aggregation area that sfincs adapter can read

panosatha:


flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -311,15 +312,41 @@ def add_green_infrastructure(
...
             polygon_file,
             geom=self.sf_model.region,
             crs=self.sf_model.crs,
         )
 
-        # Determine volume capacity of green infrastructure
+        # Make sure no multipolygons are there
+        gdf_green_infra = gdf_green_infra.explode()
 
+        # Determine volume capacity of green infrastructure
```
GundulaW, line 349: Add if-else statement if height or volume is zero. hydromt-sfincs does not seem to like this.
Add error in log file and pop up window in GUI and abort scenario run here

GundulaW (approved):
water square worked for me :)

Accidentally, I did not adjust the height so it was 0.0. The hydromt-sfincs function crashed. Can you add an error statement?

You can also make an extra issue out of it to pick this up later and merge it as is

PR #292, opened 2023-11-16
Add calculator call

dladrichem:


flood_adapt/object_model/direct_impacts.py
```diff
@@ -442,6 +477,9 @@ def _create_infometrics(self, fiat_results_df) -> Path:
         else:
             ext = ""
 
+        if self.hazard.event_mode == "risk":
```
panosatha, line 480: should this be part of the create_infometrics method? I see this as a different process so maybe it should be in the general post-process method?

panosatha:
I think the way it is implemented now the exceedance probability per object is not saved in the csv file, but only potentially used in the infometrics calculations. 

Therefore, a small part should be added that replaces the Impacts_detailed_{scenario_name}.csv in output/Scenarios/{scenario_name}/Impacts with the one including the exceedance probability.

PR #306, opened 2023-12-12
93 plot ingredients of selected projection

LuukBlom:
Also checkout to the same branch name for FloodAdapt-GUI repo and review changes.

### DIFF
--- a/flood_adapt/object_model/io/unitfulvalue.py
+++ b/flood_adapt/object_model/io/unitfulvalue.py
@@ -49,7 +49,14 @@ class VerticalReference(str, Enum):
     datum = "datum"
 
 
-class UnitfulLength(BaseModel):
+class ValueUnitPair(BaseModel):
+    value: float
+    units: str
+
+    def __str__(self):
+        return f"{self.value} ({self.units})"
+    
+class UnitfulLength(ValueUnitPair):
     value: float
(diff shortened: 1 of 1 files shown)


flood_adapt/object_model/io/unitfulvalue.py
```diff
@@ -49,7 +49,15 @@ class VerticalReference(str, Enum):
     datum = "datum"
 
 
-class UnitfulLength(BaseModel):
+class ValueUnitPair(BaseModel):
+    value: float
+    units: str
+
+    def __str__(self):
+        return f"{self.value} ({self.units})"
```
GundulaW, line 57:
```suggestion
        return f"{self.value} {self.units}"
```

GundulaW: I just removed the brackets around the unit symbol. You can accept the suggestion straightaway on github I think

GundulaW (approved):
Nice way with the ValueUnitPair. We may use that in a few more instances in the GUI.

I just removed the brackets around the units. You can accept the suggestions straightaway on github I think

GundulaW (approved):

GundulaW (approved):

panosatha (approved):

PR #309, opened 2023-12-13
Add logging

dladrichem:


flood_adapt/dbs_controller.py
```diff
@@ -90,25 +90,26 @@ def get_aggregation_areas(self) -> dict:
...
-            ).to_crs(4326)
-        else:
-            svi_map = None
-        return svi_map
+        # Read the map
+        full_path = self.static_path / path
+        if full_path.is_file():
+            return gpd.read_file(full_path, engine="pyogrio").to_crs(4326)
+
+        # If the file does not exist, return an empty GeoDataFrame
```
LuukBlom, line 111: if full_path.is_file(): will never throw an exception, it will just return true or false

So, shouldn't we raise an error or at least a warning? the exception can then be caught in get_svi_map()

flood_adapt/api/startup.py
```diff
@@ -52,9 +52,12 @@ def get_svi_map(database: IDatabase) -> GeoDataFrame:
     Returns
     -------
     GeoDataFrame
-        GeoDataFrames with the SVI per area
+        GeoDataFrames with the SVI map, None if not available
     """
-    return database.get_svi_map()
+    try:
+        return database.get_static_map(database.site.attrs.fiat.svi.geom)
+    except Exception:
```
LuukBlom, line 59: this will never reach the except block. see below

LuukBlom (changes requested):
Seems good to me! Except for two small comments/suggestions
1) the try: except: block and the if full_path.is_file() 
2) There are no tests (which is fine for now, but a TODO nevertheless)

flood_adapt/dbs_controller.py
dladrichem: True, now throwing an error. Next step would be to propper exception handling in the API

Issue #307, opened 2023-12-12
Add hurricanes to risk ensemble


GundulaW:

GundulaW:
Tasks:

- [ ] Copy spw file to event folder when created for the first time
- [ ] Before creating spw file, first check if it already exists in event folder
- [ ] Update probabilistic_event_set to include hurricanes

GundulaW:
Created separate issue for the last task, see https://github.com/Deltares-research/FloodAdapt-GUI/issues/253

PR #316, opened 2023-12-19
307 add hurricanes to risk ensemble

GundulaW:
closes https://github.com/Deltares-research/FloodAdapt/issues/307 

This contains some fixes to also run offshore simulations in a probabilistic event set. And it reuses existing spiderweb files and will not create a new one every time a hurricane is run.

There is still an outstanding issue in the frontend to remove existing spiderweb files if a hurricane event is edited. But these backend changes are not dependent on the frontend issue and can already be reviewed and merged


panosatha (approved):
seems good and I did a test run with HUGO and it worked fine.

Issue #301, opened 2023-12-07
Optimize benefits code


panosatha:
### Kind of request

Changing existing functionality

### Enhancement Description

Some parts of the code in the benefits objects are repetative (especially with repspect to the aggregated benefits). Some restracturing could make the code a bit clearer and more maintainable.

### Task list

- [x] Identify repetations and parts that can be optimized
- [x] Refactor the code accordingly

PR #317, opened 2023-12-21
benefits - include equity weighted benefit calculation

panosatha:
In order to include the equity weighted benefits I needed to do quite some changes, so I decided to tackle issue #301 at the same time, by changing the method of the aggregation areas, and using generalized static methods to not repeat the same code.

### DIFF
--- a/flood_adapt/object_model/benefit.py
+++ b/flood_adapt/object_model/benefit.py
@@ -214,7 +214,7 @@ class Benefit(IBenefit):
         implementation_cost : float
             initial costs of implementing the adaptation strategy
         annual_maint_cost : float
-            annual maintenance cost of the adaptation  strategy 
+            annual maintenance cost of the adaptation  strategy
         discount_rate : float
             yearly discount rate
 
--- a/flood_adapt/object_model/benefit.py
+++ b/flood_adapt/object_model/benefit.py
@@ -158,18 +158,18 @@ class Benefit(IBenefit):
         Parameters
         ----------
         years : list[int, int]
-            _description_
+            the current and future year for the analysis
         risk_no_measures : list[float, float]
-            _description_
+            the current and future risk value without any measures
         risk_with_strategy : list[float, float]
-            _description_
+            the current and future risk value with the strategy under investigation
         discount_rate : float
-            _description_
+            the yearly discount rate used to calculated the total benefit
 
         Returns
         -------
         pd.DataFrame
-            _description_
+            Dataframe containing the time-series of risks and benefits per year
         """
         benefits = pd.DataFrame(
             data={"risk_no_measures": np.nan, "risk_with_strategy": np.nan},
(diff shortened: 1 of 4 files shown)


panosatha:
I commited some changes in the test_database in this pull request, which **do not** need to be reviewed since we use the SVN.

tests/test_database/charleston/static/site/site_without_river.toml
LuukBlom, line 1: Are these changes on the SVN as well? I think it'd be nice to keep the SVN updated and refer to/copy the test_database from there.

tests/test_object_model/test_benefits.py
LuukBlom, line 1: I think we should refactor some of the testing in the beginning of next year. This is okay for now, but ideally, unittests should be much smaller and test one aspect of a function. At the moment you are testing the run_benefit_analysis function as a whole, and if anything goes wrong anywhere in the function, it is quite difficult/tedious to find out what exactly. 

Good practice to follow when unit testing: write 1 assert per test, and write at least 1 test per new function defined in this file (likely more than one to test all exceptions and edge cases. Each test is then named 'test_methodName_givenInputs_shouldDoSomething()', which if any test fails, immediately tells you where to look.
When we do the testing refactor, I'll show/explain the unittest template I made to reduce code duplication and keep organized when writing loads of tests) 

So TODO, writing smaller tests for each function like: 
testing 'has_run_check()'
- test_hasRunCheck_unfinishedRun_false()
- test_hasRunCheck_finishedRun_true()
- test_hasRunCheck_finishedRun_htmlCreated()
- test_hasRunCheck_invalidInputfiles_raiseFileNotFoundError()
etc for other functions

LuukBlom (commented):
The code in benefits is well structured and I haven't found a single bug! (although I havent thoroughly tested it) 
The comment about testing is meant more as information than as a suggestion to be implementated now, as we'll refactor testing later.

flood_adapt/object_model/benefit.py
```diff
@@ -145,9 +141,105 @@ def run_cost_benefit(self):
...
+        benefits = benefits.interpolate(method="linear")
+
+        # Calculate benefits
+        benefits["benefits"] = (
+            benefits["risk_no_measures"] - benefits["risk_with_strategy"]
+        )
+        # Calculate discounted benefits using the provided discount rate
+        benefits["benefits_discounted"] = benefits["benefits"] / (
+            1 + discount_rate
+        ) ** (benefits.index - benefits.index[0])
```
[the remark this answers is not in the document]
panosatha: I have now added that in the docstring

tests/test_database/charleston/static/site/site_without_river.toml
panosatha: indeed these will be updated in the SVN as well

tests/test_object_model/test_benefits.py
panosatha: Really happy to follow that approach when we do the unit test refactoring

LuukBlom (approved):

PR #321, opened 2024-01-04
Make measure creation more general

dladrichem:
Add the option of setting the database input path to a none. It doesn't change anything for FloodAdapt but for de modelbuider (where there is no database), it can now also be used to make models. And also add the thin dam and culvert options


panosatha (approved):
Looks good!

PR #328, opened 2024-01-12
corrected unit tests

panosatha:
small correction for the unit tests to make sure the database files are consistent


LuukBlom (approved):
All tests run for me now!
Given you use the most recent charleston_test and system in the svn.

PR #342, opened 2024-01-25
Update dependencies

panosatha:


panosatha:
black passes locally but fails in CI

LuukBlom (approved):
Nice that the pydantic update wasn't too much trouble.

PR #343, opened 2024-01-25
341 fix slr in event sets

GundulaW:
That was a very peculiar error where I had added n * SLR (with n being the number of synthetic or nearshore_historical events). In the probabilistic event set we have 60 nearshore_historical events so instead of adding 1ft of SLR FloodAdapt added 60ft. Hence the crazy floodmap Kathryn got.

This happened in the function hazard.add_wl(), which was actually only called once in hazard.preprocess_sfincs(). Both functions looped over all events in the event set and this is where I must have been adding SLR n times. I removed the function hazard.add_wl() since it just made it confusing and led to this error and moved its functionality directly into hazard.preprocess_sfincs(). The fix also includes an updated test to check whether SLR has been added correctly.

### DIFF
--- a/flood_adapt/object_model/hazard/hazard.py
+++ b/flood_adapt/object_model/hazard/hazard.py
@@ -226,31 +226,6 @@ class Hazard:
 
     # no write function is needed since this is only used internally
 
-    def add_wl_ts(self):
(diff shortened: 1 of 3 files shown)


panosatha (approved):
looks good

PR #349, opened 2024-02-06
Fix green infra and add tests

dladrichem:
- Fixed and added validators for the green infra model and all models it depends on
- Added tests for checking the validators
- Fixed the SFINCS adapter since the volume is always already calculated when making the measure
- Added checking the black version since it won't work if you locally run another version


flood_adapt/object_model/interface/measures.py
LuukBlom, line 1: Since validators here can be dependent on multiple fields in the class, we could write field validators & model validators in the classes where the validation is needed with `@model_validator(mode='after'):` to run the validator after all default values are set and the simpler validators (that only depend on 1 value) have run. 

Also, i think we updated to pydantic v2, but the way we have validators now is still using the deprecated v1 implementation.

https://docs.pydantic.dev/latest/concepts/validators/#field-validators
https://docs.pydantic.dev/latest/concepts/validators/#model-validators

flood_adapt/object_model/interface/measures.py
```diff
@@ -55,13 +54,33 @@ class MeasureModel(BaseModel):
     description: Optional[str] = ""
     type: Union[HazardType, ImpactType]
 
+    @validator("name")
```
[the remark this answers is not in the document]
LuukBlom: I see now that always=True is v1 syntax

https://docs.pydantic.dev/latest/concepts/validators/#validation-of-default-values

flood_adapt/object_model/interface/measures.py
```diff
@@ -138,23 +157,88 @@ class GreenInfrastructureModel(HazardMeasureModel):
...
+    ) -> Optional[UnitfulLength]:
+        if values.get("type", "") == HazardType.total_storage:
+            if height is not None:
+                raise ValueError("Height cannot be set for total storage type measures")
+            return None  # Height is not required for total storage type measures
+        elif not isinstance(height, UnitfulLength):
+            raise ValueError("Height must be a UnitfulLength")
+        elif height.value <= 0:
             raise ValueError("Height cannot be zero or negative")
         return height
```
LuukBlom, line 195: We should use @model_validators for this!

Also some nitpicking, if height is not None -> if height:

LuukBlom (changes requested):

flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -349,16 +349,9 @@ def add_green_infrastructure(
...
-            height = (
-                green_infrastructure.height.convert(UnitTypesLength("meters"))
-                * green_infrastructure.percent_area
-            )
-            volume = None
-        elif green_infrastructure.volume.value != 0.0:
-            height = None
-            volume = green_infrastructure.volume.convert(UnitTypesVolume("m3"))
+        # Volume is always already calculated and is converted to m3 for SFINCS
+        height = None
```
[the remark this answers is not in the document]
dladrichem: If I'm correct, Sfincs only requires one of them. Either the volume, or the height and it then calculates the volume itself. But I asked Gundula to review this part as well

.github/workflows/lint.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       - name: Install dependencies
         run: |
           pip install black
+      - name: Check black version
+        run: black --version
```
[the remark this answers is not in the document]
dladrichem: If we have the CICD running, then probably yes. For now, I just wanted to check what the current running version was so that I could match that version locally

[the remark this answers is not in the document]
dladrichem: Cause if we have the CICD, we also have to deliberately switch the version if needed so that all changes are in 1 PR

tests/test_object_model/interface/test_measures.py
```diff
@@ -0,0 +1,531 @@
...
+            "volume_negative",
+            "height_none",
+            "height_zero",
+            "height_negative",
+            "percent_area_none",
+            "percent_area_negative",
+            "percent_area_above_100",
+        ],
+    )
+    def test_green_infrastructure_model_greening_fails(
```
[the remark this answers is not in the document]
dladrichem: Which word?

flood_adapt/object_model/interface/measures.py
```diff
@@ -174,70 +174,44 @@ def validate_type(cls, type: HazardType, values: Any) -> HazardType:
...
+            return self
+        elif not isinstance(self.height, UnitfulHeight) or not isinstance(
+            self.percent_area, float
         ):
             raise ValueError(
-                "If `selection_type` is 'aggregation_area', then `aggregation_area_name` needs to be set."
+                "Height and percent_area needs to be set for greening type measures"
             )
-        return aggregation_area_name
+        return self
```
LuukBlom, line 201: Its not necessary, but I like explicitly using elifs for all cases and then throwing an error if the function loops through all options and doesn't find it. The field validator for type already does this check so its not required, but for readability I'd keep it.

flood_adapt/object_model/io/unitfulvalue.py
```diff
@@ -100,15 +100,26 @@ def convert(self, new_units: UnitTypesLength) -> float:
...
             new_conversion = 1.0 / 0.0254
-        elif self.units == "miles":
+        elif new_units == "miles":
             new_conversion = 1.0 / 1609.344
         else:
             ValueError("Invalid length units")
         return conversion * new_conversion * self.value
 
 
+class UnitfulHeight(UnitfulLength):
```
LuukBlom, line 110: We should make sure this is used everywhere its needed, because currently all heights are just lengths I think, so we should double check every instance of unitfullength and see if it needs to be unitfulheight

dladrichem: Added a comment to your pydantic V2 issue #350

flood_adapt/object_model/interface/measures.py
dladrichem: Technically this is already checked in the type validator but for readability, it might be better to do it in 1 statement indeed

LuukBlom (approved):
Looks good to me

flood_adapt/object_model/interface/measures.py
LuukBlom: I mean do the explicit check for the hazard.Type here 
```
...
elif self.type == HazardType.greening:
    if not isinstance(self.height, UnitfulHeight) or not isinstance(self.percent_area, float):
         ....
else:
     raise ValueError( 'invalid/unsupported hazard type')
```

LuukBlom: I mean do the explicit check for the hazard.Type at line 195.
Then we have an elif for each hazard type and throw an error if we implement a new hazardtype but forgot to update this function.
```
...
elif self.type == HazardType.greening:
    if not isinstance(self.height, UnitfulHeight) or not isinstance(self.percent_area, float):
         ....
else:
     raise ValueError( 'invalid/unsupported hazard type')
```

LuukBlom:
> It works for greening and water square but it still crashes when trying to apply a total storage to an aggregation area with the following error:
> 
> 1 validation error for GreenInfrastructureModel height Value error, Height cannot be zero or negative [type=value_error, input_value={'value': 0.0, 'units': 'meters'}, input_type=dict]

If i run the tests, all green. Also, if i open the dialog window in the gui in measures -> category(greeninfra), measure(totalstorage), area type(aggr area) -> add measure -> fill in name & description, leave total volume = 0.0 -> i get the error below. The behaviour is now: the user is notified that they cant make this particular type of measure. Which is what we wanted? 
FloodAdaptGreenInfraError
s/153174893/a0adc56f-486f-4e90-beea-e7f75bf80f10)

I do think the error message shown could use some love. Should be made clearer for the users what went wrong and what to do now without extra info about pydantic etc.

PR #353, opened 2024-02-12
Refactor projection saving and handle user uploaded shapefile

LuukBlom:


flood_adapt/dbs_controller.py
```diff
@@ -927,14 +927,28 @@ def save_projection(self, projection: IProjection) -> None:
...
+
+        projection_path = Path(self.input_path / "projections" / projection.attrs.name)
+        os.mkdir(projection_path)
+
+        # Handle user uploaded shapefile
+        if projection.attrs.socio_economic_change.new_development_shapefile is not None:
+            src_file = Path(
+                projection.attrs.socio_economic_change.new_development_shapefile
             )
+            src_file_name = src_file.name
```
dladrichem, line 939: Bit of nitpicking but this can be 1 step right? file_name = Path().name

flood_adapt/dbs_controller.py
```diff
@@ -927,14 +927,28 @@ def save_projection(self, projection: IProjection) -> None:
...
+
+            dst_path = projection_path / src_file_name
+            # OLD OPTION: dst_path = projection_path / "new_development_area.geojson"
+            # This is what it was, do we want to have all user provided shapefiles named the same?
+            # There will be only one per projection so its okay?
```
dladrichem, line 944: Maybe check this with @GundulaW. If it is indeed only possible to upload 1 file, the best option would be {projection.attrs.name}.geojson I think. Most consistent with the other stuff we save

panosatha: Indeed there is only 1 file that can have multiple polygons

flood_adapt/dbs_controller.py
```diff
@@ -938,15 +938,25 @@ def save_projection(self, projection: IProjection) -> None:
...
             )
 
+            dst_path = projection_path / "new_development_area.geojson"
+
+            gdf = gpd.read_file(src_file, engine="pyogrio")
+            with open(dst_path, "w") as f:
+                f.write(gdf.to_crs(4326).to_json(drop_id=True))
+
+        # Save the projection toml file
+        projection.save(projection_path / "new_development_area.geojson")
```
dladrichem, line 958: Moet dit dan niet projectionname.toml zijn? Die word nu nergens opgeslagen

dladrichem (approved):

PR #363, opened 2024-02-26
355 option to delete simulation folders for sfincs and fiat

panosatha:
- Added a netcdf file with the max water levels in the Flooding folder of the results for each scenario to be used instead of the actual simulation
- Implemented the deletion of the simulation folders for SFINCS and FIAT if it is chosen in the site.toml
- Changed the has_run checks for both Hazard and Direct_Impacts to ensure only post-processed output is checked.

### DIFF
--- a/flood_adapt/object_model/hazard/hazard.py
+++ b/flood_adapt/object_model/hazard/hazard.py
@@ -3,7 +3,7 @@ import os
 import shutil
 import subprocess
 from pathlib import Path
-from typing import List
+from typing import List, Union
 
 import numpy as np
 import pandas as pd
@@ -71,12 +71,9 @@ class Hazard:
         self.site = Site.load_file(
             database_input_path.parent / "static" / "site" / "site.toml"
         )
-
         self.set_simulation_paths()
 
-        self._set_sfincs_map_path(mode=self.event_mode)
-
-        self.has_run = self.sfincs_has_run_check()
+        self.has_run = self.has_run_check()
 
     @property
     def event_mode(self) -> Mode:
@@ -85,13 +82,6 @@ class Hazard:
     @event_mode.setter
     def event_mode(self, mode: Mode) -> None:
         self._mode = mode
-
-    def _set_sfincs_map_path(self, mode: Mode) -> None:
-        if mode == Mode.single_event:
-            [self.sfincs_map_path] = self.simulation_paths
-
-        elif mode == Mode.risk:
-            self.sfincs_map_path = self.results_dir
 
     def set_simulation_paths(self) -> None:
         if self._mode == Mode.single_event:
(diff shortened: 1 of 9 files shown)


flood_adapt/object_model/hazard/hazard.py
```diff
@@ -259,17 +266,14 @@ def postprocess_models(self):
...
-        #                 )  # TODO cannot remove simulation because hydromt log file is still being used
-        #             except WindowsError:
-        #                 pass
+        if not self.site.attrs.sfincs.save_simulation:
+            sim_path = self.results_dir.joinpath("simulations")
+            if os.path.exists(sim_path):
+                try:
+                    shutil.rmtree(sim_path)
+                except WindowsError:
+                    pass
```
LuukBlom, line 276: This works, but it would be better to avoid writing and deleting files where possible.
If we can add a check before writing large files to disk that are not going to be used by other parts of the codebase, we can save more time. 
Its not possible everywhere though since other models might use the created files so definitely keep this.

LuukBlom, line 276: try except blocks with only a 'pass' at the except part, are considered bad practice as they hide errors from users.

Also, OSError is the base class of WindowsError (and other OSerrors like it)

```
try:
     shutil.rmtree(sim_path)
except OSError as e_info:
     logging.warning(f"{e_info}\nCould not delete {sim_path}.")
```

tests/test_object_model/test_benefits.py
LuukBlom, line 1: General comment, why do we capitalize folder names in the output folder, but not anywhere else

flood_adapt/object_model/hazard/hazard.py
```diff
@@ -144,6 +134,23 @@ def set_simulation_paths(self) -> None:
...
+            _description_
+        """
+        self._get_flood_map_path()
+
+        has_run = False
+
+        if self.flood_map_path:
+            has_run = True
+
+        return has_run
```
LuukBlom, line 152: This is fine for now, but I feel there must be a better way to find out if a model has been run

LuukBlom (changes requested):

tests/test_object_model/test_benefits.py
panosatha: That's a good question! At some point Kathryn had made a stucture with folder naming which inlcuded that in capital, so we have been using it like that ever since.

flood_adapt/object_model/hazard/hazard.py
panosatha: That's a good one. Actually these were coded before and commented out. So I just uncommented them withtout putting much effort into checking the code again. Now I have corrected that with your suggestion.

flood_adapt/object_model/hazard/hazard.py
panosatha: I updated the method to check explicitly for file existense

tests/test_object_model/test_benefits.py
LuukBlom: Okay, might be nice to rename it at some point but its very low priority

LuukBlom (approved):
LGTM

PR #364, opened 2024-02-26
ruff and typos corrections for docs

panosatha:
the latest doc merge introduced some issues with the ruff and typos checks. 
These are now fixed


docs/index.qmd
Santonia27, line 1: The issue with removing double spaces is that then the next paragraph is not recognized and content will be written in the same line instead of the next paragraph.

Santonia27 (approved):

PR #367, opened 2024-02-28
Refactor of the save/edit/delete enz. functions for all tabs

dladrichem:


flood_adapt/dbs_classes/dbs_template.py
```diff
@@ -0,0 +1,370 @@
...
+        last_modification_date = [
+            datetime.fromtimestamp(file.stat().st_mtime) for file in paths
+        ]
+
+        objects = {
+            "path": paths,
+            "last_modification_date": last_modification_date,
+        }
+
+        return objects
```
LuukBlom, line 370: might be nice to return the paths + the last_modfication_date as a key value pair in the objects dict.
something like: `objects = {path: date for path, date in zip(paths, last_modification_date)}`

flood_adapt/dbs_classes/dbs_template.py
```diff
@@ -0,0 +1,370 @@
...
+        # Check if the object exists
+        if object_model.attrs.name not in self.list_objects()["name"]:
+            raise ValueError(
+                f"'{object_model.attrs.name}' {self._type} does not exist. You cannot edit an object that does not exist."
+            )
+
+        # Check if it is possible to delete the object by saving with overwrite. This then
+        # also covers checking whether the object is a standard object, is already used in
+        # a higher level object or is locked. If any of these are the case, it cannot be deleted.
+        self.save(object_model, overwrite=True)
```
LuukBlom, line 269: If users try to edit objects and get errors that they cannot delete it, I think it will be confusing.
Editing an object in the database is functionally the same as deleting the old and saving the new, but I think adding some duplicated code with more informative errors is okay/good here. 
We could also group the editable/delete-able checks in a function is_mutable() or something and call that twice

flood_adapt/dbs_classes/dbs_benefit.py
```diff
@@ -0,0 +1,94 @@
...
+            # Delete output if edited
+            output_path = (
+                self._database.input_path.parent
+                / "output"
+                / "Benefits"
+                / benefit.attrs.name
+            )
+
+            if output_path.exists():
+                shutil.rmtree(output_path, ignore_errors=True)
```
LuukBlom, line 94: I didnt know you could do try except else, seems strange but it works so nice one!

flood_adapt/dbs_classes/dbs_template.py
```diff
@@ -0,0 +1,370 @@
...
+        if model_object is None and name is None:
+            raise ValueError("Either model_object or name must be provided.")
+
+        # If only name is provided, get the model_object
+        if model_object is None:
+            model_object = self.get(name)
+
+        # Set the lock and save the object
+        model_object.attrs.lock_count += 1
+        self.save(model_object, overwrite=True)
```
LuukBlom, line 83: I think there might be a bug here.
If we start with an obj that is already locked, and want to use it in another higher level obj:
self.save(obj, overwrite=True) gets called, which calls self.delete(obj)
which then will raise a ValueError at L259

flood_adapt/dbs_classes/dbs_benefit.py
dladrichem: Now that I look at it again, I also don't really know what my idea was here. The only thing it adds now is that it still tries to save it when you get an error other than a ValueError. Which doesn't even feel right haha

flood_adapt/dbs_classes/dbs_template.py
dladrichem: If you edit, you then still need to delete the files as well. So that would mean that you have to add all the code in the delete function to this new function, or duplicate the code to delete. I now added some text to the error which should make it a bit more descriptive. Also, in the gui, when this error is caught, you can also change the error there based on the function it is called from

flood_adapt/dbs_classes/dbs_benefit.py
dladrichem: And the locked behaviour can be removed. Requirements now are that if something is in a scenario, it cannot be deleted anyway. So this functionality is obsolete

flood_adapt/dbs_classes/dbs_template.py
dladrichem: The idea of the function is that you can easilly get a list of all paths and in this case also the dates (not sure if we use them anywhere). If you look at where this function is called, this dict will then be extended with a list of names, descriptions e.d.

flood_adapt/dbs_classes/dbs_template.py
```diff
@@ -0,0 +1,271 @@
...
+            Raise error if object to be deleted is already in use.
+        """
+        # Check if the object is a standard object. If it is, raise an error
+        if self._check_standard_objects(name):
+            raise ValueError(
+                f"'{name}' cannot be deleted/modified since it is a standard {self._type}."
+            )
+
+        # Check if object is used in a higher level object. If it is, raise an error
+        if used_in := self._check_higher_level_usage(name):
```
LuukBlom, line 197: cool way to do if statements, had to look it up to see if it wasnt a typo, but nice one

LuukBlom (approved):

PR #371, opened 2024-03-01
Docs getting started

kathrynroscoe:
This was mostly a test! Just added one sentence.


config.toml
Santonia27, line 1: I think this should not be changed. Please change it back to the original

kathrynroscoe: I changed this back now!

PR #388, opened 2024-03-13
Docs projections climate

kathrynroscoe:
Added information about climate projections


config.toml
Santonia27, line 1: Is this file changed and pushed on purpose?

Santonia27, line 1: Is this file changed and pushed on purpose?

docs/_static/images/Projection_window.png
Santonia27, line 1: Maybe we can convert the images into .svg files since they take less space and are resolution-independent no matter which scale. 
There should be online tools to easily convert these

docs/_static/images/green_infra_CRC.png
Santonia27, line 1: same comment as other image

docs/_static/images/projections_stormFreq.png
Santonia27, line 1: same comment as other image

Santonia27 (changes requested):

Santonia27 (approved):

PR #396, opened 2024-03-27
394 tide gauge download from local files

GundulaW:


GundulaW:
I tested it for the cork database and checked that it is still compatible with the charleston_training database.
I have not included unit tests

LuukBlom (approved):
Looks good to me

PR #417, opened 2024-04-09
Add validators

dladrichem:
Validator to exclude unwanted characters from the name. Apparently the //// is needed to actually exclude a backslash. This is because it is also used as the escape character

### DIFF
--- a/flood_adapt/object_model/interface/benefits.py
+++ b/flood_adapt/object_model/interface/benefits.py
@@ -14,7 +14,7 @@ class CurrentSituationModel(BaseModel):
 class BenefitModel(BaseModel):
     """BaseModel describing the expected variables and data types of a Benefit analysis object"""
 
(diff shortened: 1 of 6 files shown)


LuukBlom (changes requested):
Looks good, I think it would be nice if we put the pattern string in a var and use that throughout the code so we can easily change all of the regexpatterns if we want.
We could also write a field validator and re-use it https://docs.pydantic.dev/2.0/usage/validators/#reuse-validators

dladrichem:
> Looks good, I think it would be nice if we put the pattern string in a var and use that throughout the code so we can easily change all of the regexpatterns if we want. We could also write a field validator and re-use it https://docs.pydantic.dev/2.0/usage/validators/#reuse-validators

What makes it a bit weird is that all models are in different files. What we could do is make our own version of the base model that always includes the name (and maybe description since those 2 are the only ones used everywhere) or make a seperate utils file for standard validators, but I'm not 100% sure if that's worth it if is just for checking the name

PR #427, opened 2024-04-24
using z-value in floodwall from shp file 

GundulaW:
and also converting possible  MultiLineStrings to LineStrings to work with hydromt-sfincs

At the moment, this requires the shape file to have a z-column. Otherwise, the uniform elevation value from the GUI will be taken. I am trying to make this neater in the GUI by allowing the user to specify whether a unifrom value should be used for all floodwalls or which column should eb taken from the shape file. But I have not succeeded yet. This will be a future PR in the GUI

### DIFF
--- a/flood_adapt/integrator/sfincs_adapter.py
+++ b/flood_adapt/integrator/sfincs_adapter.py
@@ -298,7 +298,15 @@ class SfincsAdapter:
         ):  # The column name and the choice if uniform value or from shape file should be an option in the GUI
             gdf_floodwall["z"] = floodwall.elevation.convert(UnitTypesLength("meters"))
         else:
-            heights = [float(UnitfulLength(value=float(height),units=self.site.attrs.gui.default_length_units).convert(UnitTypesLength("meters"))) for height in gdf_floodwall["z"]]
+            heights = [
+                float(
+                    UnitfulLength(
+                        value=float(height),
+                        units=self.site.attrs.gui.default_length_units,
+                    ).convert(UnitTypesLength("meters"))
+                )
+                for height in gdf_floodwall["z"]
+            ]
             gdf_floodwall["z"] = heights
         # par1 is the overflow coefficient for weirs
         gdf_floodwall["par1"] = 0.6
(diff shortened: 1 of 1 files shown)


flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -291,7 +291,10 @@ def add_floodwall(self, floodwall: FloodWallModel, measure_path=Path):
 
         # Add floodwall attributes to geodataframe
         gdf_floodwall["name"] = floodwall.name
-        gdf_floodwall["z"] = floodwall.elevation.convert(UnitTypesLength("meters"))
+        if (gdf_floodwall.geometry.type == "MultiLineString").any():
```
panosatha, line 294: I don't think that a MultiLineString can have multiple z values. It is still a single feature, but with multiple lines, and only a single z value can be connected to it.

I think you would like to have multiple features with different z values attributed to them.

flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -291,7 +291,10 @@ def add_floodwall(self, floodwall: FloodWallModel, measure_path=Path):
...
+            gdf_floodwall = gdf_floodwall.explode()
+        if "z" not in gdf_floodwall.columns:
+            gdf_floodwall["z"] = floodwall.elevation.convert(UnitTypesLength("meters"))
```
panosatha, line 297: This would mean that if "z" is already in the columns you will not convert the units.

flood_adapt/integrator/sfincs_adapter.py
```diff
@@ -291,7 +291,10 @@ def add_floodwall(self, floodwall: FloodWallModel, measure_path=Path):
...
         gdf_floodwall["par1"] = 0.6
```
panosatha, line 298: Even thought this is not a new addition I am curious on what this par1 is :)

flood_adapt/integrator/sfincs_adapter.py
GundulaW: Yes, that's true. I am turning it into linestrings because hydromt-sfincs cannot handle MultLineStrings. These LineStrings will simply have the same z-value as the initial MultLineSrting

flood_adapt/integrator/sfincs_adapter.py
GundulaW: Good one, changed this.

flood_adapt/integrator/sfincs_adapter.py
GundulaW: I think we spoke about it

panosatha (approved):

PR #446, opened 2024-05-17
Tipping points

dumontgoulart:
Pulling tipping points into the main branch. For now we have only tipping points for the backend. 

Tipping points takes user inputs for specific impact metrics and calculates iteratively if simulations reach them. Main current limitations: For now it's only designed for sea level rise scenarios. No postprocessing done yet -> no html and plots available

This pull request is to be reviewed.


flood_adapt/object_model/tipping_point.py
```diff
@@ -0,0 +1,328 @@
...
+        self.scenarios = df.astype(
+            dtype={
+                "event": "str",
+                "projection": "str",
+                "strategy": "str",
+                "scenario created": "str",
+                "scenario run": bool,
+            }
+        )
+        return self.scenarios
```
LuukBlom, line 255: Im not convinced this is good behaviour. We are doing too much database related stuff in this class, which will produce headaches! Instead, this class should have standalone functions that the database can call after constructing the target_path (like load_file(target_path) and save(target_path) in other classes). This will decouple the database and tipping point and will make refactoring/testing/life in general much better as the tipping point now takes in any path!

A Tipping point object should be able to exist without a database object, which isnt possible now.
I think that creating a folder next to the tippingpoint.toml, that contains the input folders for all scenarios is nicer. The database can then just loop over whatever is in the tippingpoint input folder, and compare the tipping point scenarios to the ones stored in the database itself to see if the output can be copied, construct the result path, and save results) 

So, defer the running/checking/saving of the tippingpoint scenarios to the database, similar to Database.run_scenario(), which can then can call tippingpoint.evaluate(), save results etc.

flood_adapt/object_model/tipping_point.py
```diff
@@ -0,0 +1,328 @@
...
+        """create risk event from toml file"""
+
+        obj = TippingPoint(database_input_path)
+        with open(filepath, mode="rb") as fp:
+            toml = tomli.load(fp)
+        obj.attrs = TipPointModel.model_validate(toml)
+        return obj
+
+    def load_dict(
+        dct: Union[str, Path], database_input_path: Union[str, os.PathLike]
```
LuukBlom, line 270: Every load_dict function in the code base has one argument which is the dict you are trying to load. So we definitely want to keep consistent with that. (in addition to my earlier comments)

LuukBlom (commented):
Nice job, but as always there is room for improvement!

We are doing too much database related stuff in this class, which will produce headaches! Instead, this class should have standalone functions that the database can call after constructing the target_path (like load_file(target_path) and save(target_path) in other classes). This will decouple the database and tipping point and will make refactoring/testing/life in general much better as the tipping point now takes in any path!

A Tipping point object should be able to exist without a database object, which isnt possible now.
I think that creating a folder next to the tippingpoint.toml, that contains the input folders for all scenarios is nicer. The database can then just loop over whatever is in the tippingpoint input folder, and compare the tipping point scenarios to the ones stored in the database itself to see if the output can be copied, construct the result path, and save results) 

So, defer the running/checking/saving of the tippingpoint scenarios to the database class, similar to Database.run_scenario(), which can then can call tippingpoint.evaluate(), save results etc.

flood_adapt/object_model/interface/tipping_points.py
```diff
@@ -0,0 +1,101 @@
...
+    """BaseModel describing the expected variables and data types of a Tipping Point analysis object"""
+
+    name: str
+    description: Optional[str] = ""
+    strategy: str
+    event_set: str
+    projection: str
+    sealevelrise: list[float]  # could be a numpy array too
+    tipping_point_metric: list[tuple[TippingPointMetrics, float, TippingPointOperator]]
+    status: Optional[TippingPointStatus] = TippingPointStatus.not_reached
```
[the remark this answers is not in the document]
dumontgoulart: Done!

flood_adapt/object_model/tipping_point.py
```diff
@@ -0,0 +1,328 @@
...
+- Results for each scenario indicating if a tipping point is reached.
+- Compiled results are saved as CSV files.
+"""
+
+
+class TippingPoint(ITipPoint):
+    """Class holding all information related to tipping points analysis"""
+
+    attrs: TipPointModel
+    database_input_path: Union[str, os.PathLike]
```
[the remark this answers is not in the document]
dumontgoulart: Done!

flood_adapt/object_model/tipping_point.py
```diff
@@ -0,0 +1,328 @@
...
+                "name": f"slr_" + str(slr).replace(".", ""),
+                "event": self.attrs.event_set,
+                # get a string from slr removing the dot
+                "projection": self.attrs.projection
+                + "_slr"
+                + str(slr).replace(".", ""),
+                "strategy": self.attrs.strategy,
+            }
+            for slr in self.attrs.sealevelrise
+        }
```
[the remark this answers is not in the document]
dumontgoulart: done!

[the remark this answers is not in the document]
LuukBlom: Im sorry my previous example was not great.
allow me to retry:

f-strings allow you to embed expressions and variables inside string literals, using curly braces {}, which is much more readable and less bug prone than string concatenation (str + str).

f-strings fully elimminate the need for str + str syntax!
```
x = str(slr).replace(".", "")

scenarios = 
{
  f"slr_{x}":
    {
        "name": f"slr_{x}",
        "projection": f"{self.attrs.projection}_slr{x}"
    }
}
```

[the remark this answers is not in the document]
dumontgoulart: done!

tests/test_object_model/test_tipping_points.py
```diff
@@ -0,0 +1,92 @@
...
+@pytest.fixture
+def setup_database():
+    # Mock the database setup or configure a test database
+    read_database(
+        rf"C:\\Users\\morenodu\\OneDrive - Stichting Deltares\\Documents\\GitHub\\FloodAdapt-Database",
+        "charleston_full",
+    )
+    set_system_folder(
+        rf"C:\\Users\\morenodu\\OneDrive - Stichting Deltares\\Documents\\GitHub\\FloodAdapt-Database\\system"
+    )
```
LuukBlom, line 22: This already exists in conftest.py! 

You can access the test database like this (the fixtures you can use are at the bottom of conftest.py):
```
def test_tipping_point_creation(test_db):
  test_db.events.get()
```

tests/test_object_model/test_tipping_points.py
```diff
@@ -0,0 +1,92 @@
...
+        "description": "",
+        "event_set": "extreme12ft",
+        "strategy": "no_measures",
+        "projection": "current",
+        "sealevelrise": [0.5, 1.0, 1.5],
+        "tipping_point_metric": [
+            ("FloodedAll", 34195.0, "greater"),
+            ("FullyFloodedRoads", 2000, "greater"),
+        ],
+    }
```
LuukBlom, line 38: You can put the dict in a fixture and then use it in more than 1 test:

```
@pytest.fixture()
def tp_dict():
      tp_dict = {
        "name": "tipping_point_test",
        "description": "",
        "event_set": "extreme12ft",
        "strategy": "no_measures",
        "projection": "current",
        "sealevelrise": [0.5, 1.0, 1.5],
        "tipping_point_metric": [
            ("FloodedAll", 34195.0, "greater"),
            ("FullyFloodedRoads", 2000, "greater"),
        ],
    }
  return tp_dict
  
test_xx_xx(test_db, tp_dict):
  test_point = TippingPoint.load_dict(tp_dict)
  # the rest of the test
  
```

tests/test_object_model/test_tipping_points.py
LuukBlom, line 1: Nice start with the tests, here are some pointers to write better tests:

- keep them small (test one thing, can be with multiple asserts, but one aspect of the function/module)
- use a structure like Arange, Act, Assert to setup your tests (already doing that, so keep it up!)
- use descriptive names: `def test_<function_name>_<context>_<expected_result>`, so for example: `test_createTippingPoints_scenariosAlreadyExist_notDuplicated`. This ensures that when a test fails in the future, we know exactly where to go in the code to fix it.
- test edge cases! not only the happy path where everything goes as expected. testing that an error is raised when calling a function is perfectly valid!

LuukBlom (commented):
Nice work! I left some specific comments on the test file.

Please have a look at conftest.py and read the info to see how/what you can use from there!
Main points are to cleanup the tipping point class (wait for dbs_tp) & test edge cases (handle weird inputs / check error are raised correctly).

tests/test_object_model/test_tipping_points.py
dumontgoulart: Thank you! Done!

dumontgoulart:
Hi @LuukBlom , I updated the test_tipping_points.py file with some extra tests for some of the functions. Let me know if you think there are some things that need refactoring or some functions/test I'm missing :).

flood_adapt/object_model/tipping_point.py
```diff
@@ -0,0 +1,338 @@
...
+            tp_path.mkdir(parents=True)
+
+        tp_results = pd.DataFrame.from_dict(self.scenarios, orient="index").reset_index(
+            drop=True
+        )
+        tp_results["sea level"] = [
+            float(i) / 10 for i in self.attrs.sealevelrise
+        ]  # TODO: hardcoded for SLR in Meters, fix later
+        # calculate the sea level at the tipping point threshold
+        # TODO: not super clear but maybe talk to @luuk about how to better show the table
```
dumontgoulart, line 169: @LuukBlom, the way we thought to have the csv file is not really aligned to how they use the tipping points in the pathway generator. I'm not sure if they are using in the best way anyway, so I added some information but in a different format. Would you have any thoughts?

flood_adapt/object_model/tipping_point.py
```diff
@@ -0,0 +1,338 @@
...
+            x = valid_data["sea level"].values
+            y = valid_data[f"{metric_name}_value"].values
+
+            # Create a linear interpolation function
+            interpolation_function = interp1d(y, x, fill_value="extrapolate")
+
+            tp_results[f"{metric_name}_ATP"] = interpolation_function(threshold)
+
+    # create a function that plots the results from tp_path / "tipping_point_results.csv" against the SLR values
+    def plot_results(self):
```
dumontgoulart, line 235: the plot function is quite simple for now: have a linear function connecting different combinations of damage and sea level, and comparing with the tipping point thresholds. It generates a different plot for each tipping point metric. Any thoughts?

flood_adapt/object_model/tipping_point.py
dumontgoulart: Also something important to consider, but I still have to think better on what is the best solution: they use different strategies in the same place (as seen in the table), whereas we have a single strategy for all the scenarios. Should we adapt the tipping point object or make a wrapper to combine all these in plots and tables?

flood_adapt/object_model/tipping_point.py
```diff
@@ -0,0 +1,338 @@
...
+    )
+    set_system_folder(
+        r"C:\\Users\\morenodu\\OneDrive - Stichting Deltares\\Documents\\GitHub\\Database\\system"
+    )
+
+    tp_dict = {
+        "name": "tipping_point_test",
+        "description": "",
+        "event_set": "extreme12ft",
+        "strategy": "no_measures",
```
dumontgoulart, line 323: @LuukBlom, also something important to consider, but I still have to think better on what is the best solution: they use different strategies in the same place (as seen in the table), whereas we have a single strategy for all the scenarios. Should we adapt the tipping point object or make a wrapper to combine all these in plots and tables?

flood_adapt/object_model/tipping_point.py
dumontgoulart: image

flood_adapt/dbs_classes/dbs_tipping_point.py
```diff
@@ -0,0 +1,71 @@
+import shutil
+
+from flood_adapt.dbs_classes.dbs_template import DbsTemplate
+from flood_adapt.object_model.interface.tipping_points import ITipPoint
+from flood_adapt.object_model.tipping_point import TippingPoint
+
+
```
dumontgoulart, line 7: I based on the other codes and have not added other functionalities. Maybe we need to add the plots here?

flood_adapt/api/tipping_points.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any
+
+from flood_adapt.dbs_controller import Database
+from flood_adapt.object_model.interface.tipping_points import ITipPoint
+from flood_adapt.object_model.tipping_point import TippingPoint
+
```
dumontgoulart, line 6: All based on the existing code, changing paths and variables.

flood_adapt/api/tipping_points.py
```diff
@@ -0,0 +1,42 @@
...
+
+def edit_tipping_point(tipping_point: ITipPoint) -> None:
+    Database().tipping_points.edit(tipping_point)
+
+
+def delete_tipping_point(name: str) -> None:
+    Database().tipping_points.delete(name)
+
+
```
dumontgoulart, line 32: agreed with Luuk that for now we run the TP functions directly from the API

PR #449, opened 2024-05-24
Restructure static API functions and add results caching

dladrichem:
- Create seperate class for static database functions (similar to the other database classes)
- Add caching for the results of the static database functions
- Change the api_startup to api_static
- Fix small bug in UnitfulLength


LuukBlom (commented):
Test refactor wrt database singleton is already in main.
So merge main into this one to resolve/prevent conflicts?

PR #462, opened 2024-06-12
Updates in site.toml to clear not-used attributes and update some vairables

panosatha:
Various fixes to avoid having optional configuration as mandatory in the site.toml:

- remove reference_water_level.reference as it is was not used
- remove sfincs.ambient_pressure
- obs_point isn ow optional and FA should not break when you have no observation points
- renamed obs_station to tide_gauge, and removed reference water levels from it
- tide_gauge has now a source attr where the either the cht_observation source is defined or "file" to use a downloaded file
- renamed risk.flooding_threshold to flooding_frequency.threshold
- river is now optional 
- slr scenarios are now better defined in the site.toml
- offshore model and cyclone tracks not mandatory


GundulaW (commented):
Backend looks fine, but some things in the GUI broke, see GUI PR.

GundulaW (approved):
All working for me now

PR #478, opened 2024-07-04
380 remove useless statements from log file

dladrichem:


flood_adapt/integrator/fiat_adapter.py
```diff
@@ -25,7 +26,7 @@ class FiatAdapter:
     def __init__(self, model_root: str, database_path: str) -> None:
         """Load FIAT model based on a root directory."""
         # Load FIAT template
-        self._logger = FloodAdaptLogging.getLogger(__name__)
+        self._logger = FloodAdaptLogging.getLogger(__name__, level=logging.ERROR)
         self.fiat_model = FiatModel(root=model_root, mode="r", logger=self._logger)
```
LuukBlom, line 30: I like the logger for the console having reduced messages, however, would it not be nice to write all logs to disk for debugging purposes?
What do you think about making 2 loggers, 1 for the fiat adapter with only high level info and one for the FiatModel that creates the hydromt.log.

dladrichem: In that case, it's better to fix the closing of loggers in the hydromt's

LuukBlom: For hydromt loggin, you are expected to create and close it yourself. Which is why you have to pass it to constructors of hydromt model plugins

LuukBlom (approved):

PR #489, opened 2024-07-15
add pixi.toml and lockfile

Jaapel:
Some projects are already working with pixi.
It is a nice environment manager for conda / mamba packages that is very fast in resolving and installing packages. For now I would like to use it to install the environment.

Did some testing and messing around to get pixi and submodules working together.

The new installation workflow using pixi is described in the readme, so please follow instructions there and let me know if you run into issues.

Should be merged at the same time as [frontend-PR](https://github.com/Deltares-research/FloodAdapt-GUI/pull/444).


pixi.toml
```diff
@@ -0,0 +1,49 @@
...
+tomli-w = "*"
+
+[pypi-dependencies]
+cht-cyclones = { git = "https://github.com/deltares/cht_cyclones.git" }
+cht-meteo = "*"
+cht-observations =  { git = "https://github.com/deltares/cht_observations.git" }
+cht-tide = "*"
+fiat-toolbox = "*"
+hydromt-fiat = { git = "https://github.com/deltares/hydromt_fiat.git", rev = "2c1f0556103edeedcd5aec6b59ff1e3554061df1" }
+hydromt-sfincs = { git = "https://github.com/deltares/hydromt_sfincs.git", rev = "97c872499cbcaec6092931e4f0ddea1cb57779ca" }
```
savente93, line 49: don't some of these also need a branch specified?

LuukBlom (commented):
Looks logical, I do have some concerns about the pinning to commit hashes, but I assume there is a solution within pixi for that

savente93:
> Looks logical, I do have some concerns about the pinning to commit hashes, but I assume there is a solution within pixi for that

Not sure about pinned commits, but in core we would just have a github action that would periodically update dependencies, I'm sure this can be made to work with the revisions as well, though I do also think we want to use branches instead of pinned commits no?

Jaapel:
You cannot refer a branch in pixi according to the options I find in the docs: https://pixi.sh/v0.27.1/reference/project_configuration/#git

So I pinned commits. I can change to the url `#branchname` that `git` understands, let me try it out.

savente93 (approved):
:D

savente93:
@LuukBlom think we can merge? if so can you approve? because I did the merge with main I can't be the reviewer apparently

savente93:
That makes sense, I'm in favor of making sure it works well first as well, I just didn't realize there were things left. 

I can look into updating the commit hashes, but I've also been working on removing the need to use a branch on our dependencies and making some good progress. So I think if you agree it's better to just wait until we don't have to use the branches anymore. Regardless I do agree we shouldn't merge this before we have either of those solutions

LuukBlom:
Actually @savente93, using a fresh install of pixi, cloning the repo, and then the task tests: `pixi run tests` is succesfull, as well as the user guide task: `pixi run docs`.

So I think it'd be good to just merge this to get pixi in there (and stop additional scope creep) and then open issues for updating the hashes automatically and getting the api docs & build exe/installer tasks to work.

Im testing it now on wsl to see if the tests are green there as well.

What do you say?

savente93:
If it works already, then I agree. I think it will be really nice to have this as a starting point

LuukBlom:
Closing and reopening to hopefully let teamcity see this branch

LuukBlom:
replaced by https://github.com/Deltares-research/FloodAdapt/pull/567/

PR #491, opened 2024-07-15
made infographic creation not mandatory

panosatha:


flood_adapt/object_model/direct_impacts.py
```diff
@@ -517,7 +518,11 @@ def _create_infometrics(self, fiat_results_df) -> Path:
             ext = ""
 
         # Get options for metric configurations
-        metric_types = ["mandatory", "infographic", "additional"]
```
dladrichem, line 520: If we're going to make more and more things optional (i.e. the tipping points could also be optional), a better way to then do this is by building up the list like:
```
metrics_types = ['mandatory", "additional"]
if site.infometrics:
  metrics_types += 'infometrics'
if site.tipping_points: Not yet but at some point
  metrics_types += 'tipping_points'
```

flood_adapt/object_model/interface/site.py
```diff
@@ -201,6 +201,7 @@ class FiatModel(BaseModel):
     new_development_file_name: Optional[str] = None
     save_simulation: Optional[bool] = False
     svi: Optional[SVIModel] = None
+    infographics: Optional[bool] = True
```
dladrichem, line 204: Should the default be True?

panosatha: that is a good point. Probably it makes more sense for False.

flood_adapt/object_model/direct_impacts.py
panosatha: I will change it like this

dladrichem (approved):

PR #498, opened 2024-07-23
Crude start to a cleanup script

dladrichem:
Very crude start to a cleanup script. Now it just zips failed runs and puts them in a temp folder so they can be removed after a certain time


LuukBlom:
Closing this as its already implemented to optionally delete crashed runs

PR #500, opened 2024-07-29
feat(misc): automatic API doc generation

LuukBlom:
Since we don't really have a good other place to put it, we decided to just use `quartodoc` and add it to the existing documentation. The need to list all the elements included by hand in the `_quarto.yml` is unfortunate but it's a consequence of the import structure we have.


savente93:
going to continue monday, but due to checkout and premission reasons the current game plan is: 

1. have the wiki repo as submodule so the ci chan check it out
2. when needed build into the repo
3. push the changes, 
4. profit

savente93:
blocked until a new palce for the docs is agreed upon given that the github wiki is not suitable

savente93:
Without a good place to put it I decided to just add an additional section to the existing documentation

_environment.yml
LuukBlom, line 1: remove this pls

dev_docs/source/conf.py
```diff
@@ -9,23 +9,39 @@
...
+    version = None
+    with open(ROOT_PATH / "flood_adapt" / "__init__.py", "r") as f:
+        for line in f:
+            if line.startswith("__version__"):
+                version = line.split("=")[1].strip().strip('"')
+    assert version, "Could not find version in __init__.py"
+    return version
+
+
+release = get_floodadapt_version()
```
LuukBlom, line 44: Nitpicky and up to you, but why not `from flood_adapt import __version___`, seems cleaner imo. Or is flood_adapt not installed in the docs environment?

docs/_quarto.yml
```diff
@@ -25,11 +24,45 @@ metadata-files:
...
+        - name: dbs_classes.dbs_strategy.DbsStrategy
+        - name: dbs_classes.dbs_template.DbsTemplate
+        - name: dbs_controller.Database
+        - name: object_model.benefit.Benefit
+        - name: object_model.direct_impacts.DirectImpacts
+        - name: object_model.scenario.Scenario
+        - name: object_model.site.Site
+        - name: object_model.strategy.Strategy
+        - name: FloodAdaptLogging
+  package: flood_adapt
```
LuukBlom, line 60: Is there a way this can be generated from source code? 
This is 100% something people will forget to update if they add new files / classes

LuukBlom (commented):
Any specific reason the imports in the `__init__.py` files are added? Not that I disagree, I just want to understand

docs/_quarto.yml
savente93: in theory? yes. As is? no. This is also to answer the question you put in your review comment. This is not what it does exactly, but what it seems `quartodoc` does to build it's links is `from <package> import *` to figure out what to document. Usually this works quite well since submodules usually re-export their public parts so by just importing everything from the root, you can get the important things. The reason that this doesn't work for us, is that we don't import our submodules from the root, presumably because of circular imports which I ran into immediately when I tried that. This was the best compromise I could find

dev_docs/source/conf.py
savente93: whoops, this is part of an old solution and not actually used anymore. so I'll remove it

environment/_environment.yml
LuukBlom, line 1: this was actually the correct env.yml that is used by the make_env script. The one you should remove is in the root level of the repo and contains the absolute paths (put there by me so sorry)

_environment.yml
LuukBlom, line 1: remove this file pls

LuukBlom (commented):
I like it, could you also provide some isntructions on how to build the docs locally? Just so that it is written down and we dont have to figure it out again in half a year

docs/_quarto.yml
LuukBlom: Yeah I remember there being problems with that. Ok it is what it is then. Also since its the api, these things shouldnt change that often

savente93 (approved):

PR #503, opened 2024-08-01
Update workflows

savente93:
Summary of changes: 

1. updated all the used action versions to the latest vesion, as some were really quite out of date, and made any changes necessary to fix whatever broke
2. removed black in favour of ruff formatting as ruff formatting implements the same formatting rules as black but runs at least an order of magnitude faster
3. did the same two things for the pre-commit config
4. ran pre-commit on all files 
5. removed the manual linting steps in favor of just using the pre-commit action which is both faster (does better caching) and ensures both local and ci checks are consistent 

I know because of step 4 the number of changed files is kinda ridiculous but no meaningful differences were made, only the automated fixes from the pre-commit hooks so no need to review them individually 

TODO: 
1. remove required status of removed workflows if this is accepted.


.pre-commit-config.yaml
LuukBlom, line 1: Make sure the pinned versions for ruff/typos/black are the same as in the pyproject.toml file. If these are different, the pre-commit people have locally can format differently then the one on github, making it fail 100% of the time.

pyproject.toml
LuukBlom, line 1: remove black from dependencies if we are not using it anymore

savente93:
@LuukBlom  fixed your comments, should be good now

LuukBlom (approved):

LuukBlom (approved):

PR #514, opened 2024-08-21
Don't offsett water levels in synthetic events

### LINKED ISSUE #410: Water level offset should not be included in synthetic events
### FloodAdapt version checks

### Reproducible Example

if you make an event with an offshore model, floodadapt will always add the water level offset to the overland surge & tide, even when you have specified a synthetic overland model (which should not have this addition)

### Current behaviour

always add water level offset

### Desired behaviour

Offshore model with historic gauges run?
Needs water level offset since the historic gauges need to be corrected for slr.
Otherwise it does not need it and will just use whatever you have provided in the overland model

### Task list

events refactor

### Additional context

_No response_

savente93:
Fixes #410 

Not really sure how to test this, but Luuk told me this was about what was necessary. This is also already fixed in the events refactor so this is just a temporary fix until we can land that.


savente93:
Tests are failing spectacularly so it seems I will have to do more investigation yet

LuukBlom (approved):

PR #517, opened 2024-08-23
feat: Improve copy method to DbsTemplate class for duplicating object…

LuukBlom:
…s, but excluding some file suffixes. (toml by default, .spw for events)


panosatha (approved):
Looks good!

PR #526, opened 2024-08-28
bug fix in benefit scenarios creation

panosatha:


flood_adapt/dbs_controller.py
```diff
@@ -666,8 +666,12 @@ def create_benefit_scenarios(self, benefit: IBenefit) -> None:
...
 
                 scenario_obj = Scenario.load_dict(scenario_dict, self.input_path)
-
-                self._scenarios.save(scenario_obj)
+                # Check if scenario already exists
+                if (
+                    scenario_obj.attrs.name
+                    not in self.scenarios._get_object_list()["name"]
+                ):
+                    self._scenarios.save(scenario_obj)
```
LuukBlom, line 674: This check (among other checks) is in dbsTemplate.save(), which is what is called for saving scenarios. What is the idea of adding it here?

Also it is better practice to use the property `self.scenarios` for the scenarios instead of accessing directly the private attribute (denoted by starting with `_`)

Personally, I'd add a try: except ValueError around the save call and check that the raised error is the error raised at L140 in dbsTemplate. (by string comparison or we can write some custom exception classes for the database maybe?) 

```
try:
  self.scenarios.save(scn_obj)
except ValueError as e:
  if not 'name is already used' in str(e): 
    # some other error was raised, so we re-raise it
    raise e
  # otherwise, t already exists and we dont need to save it (are we sure about that?), we can just continue
```

panosatha: done

LuukBlom (approved):

Issue #485, opened 2024-07-11
Install via Pixi


Jaapel:
### Kind of request

Adding new functionality

### Enhancement Description

Installing via the python3.10-win-amd wheels only is difficult for people with different environments (and for sure the cloud!).
Using pixi (also used in other projects / products) can help make this easier for developers and speed up the ci.

### Task list

- [x] write pixi.toml with requirements, versions
- [ ] create important tasks in pixi
- [ ] (optional) use these tasks in the ci

### Use case

Installing FloodAdapt backend dependencies in a non-windows environment.

kathrynroscoe:
This is going to make it easier for people to install FA (python version) and to have it running online

savente93:
Pixi does not support authenticated/private repositories so until the FA Quadtree functionality from the SFINCS team is migrated to the public repository this is blocked

PR #527, opened 2024-08-29
Running FloodAdapt on Linux

Jaapel:
In order to run on linux, a few changes were made:
- Needed to update how fiat was built on linux: https://github.com/Deltares/Delft-FIAT/pull/126
- Build SFINCS on linux using: https://github.com/Deltares/SFINCS/blob/main/source/Dockerfile.cpu
- Needed to change some case-sensitive filepaths, for example `exposure` vs. `Exposure`
- Needed to make the executable names of `fiat` and `sfincs` configurable ( no `.exe` outside of windows)
- Changes the database to use lowercase folder names where the code also had these

To do configuration I added `pydantic-settings` to our dependencies (maintained by `pydantic` maintainers, which is already in our dependencies.)


Jaapel:
@LuukBlom can you take a look? We should chat on how we will deal with case-sensitive filepaths (with capitalization or without)

flood_adapt/config.py
LuukBlom, line 1: This is great! nice job

flood_adapt/object_model/benefit.py
```diff
@@ -202,7 +202,7 @@ def cba(self):
         scenarios = self.scenarios.copy(deep=True)
         scenarios["EAD"] = None
 
-        results_path = self.database_input_path.parent.joinpath("output", "Scenarios")
+        results_path = self.database_input_path.parent.joinpath("output", "scenarios")
```
LuukBlom, line 205: This is the old way of getting paths. We have a function in the database controller to get the paths so at least all of the paths are constructed in the same way. 

see [dbsTemplate::get_database_path()](https://github.com/Deltares-research/FloodAdapt/blob/41a0a681aa6641ef856b5ebf8f0cf797cc21447b/flood_adapt/dbs_classes/dbs_template.py#L252)

The only problem is that we need to have the database accessible eveywhere, so its not a perfect solution by any means, but at least it more stable than hardcoding the strings and paths throughout the code

flood_adapt/config.py
```diff
@@ -1,167 +1,85 @@
...
+    def database_path(self) -> Path:
+        return self.database_root / self.database_name
+
+    @model_validator(mode="before")
+    def set_defaults_if_missing(cls, data: Any) -> "Settings":
+        if isinstance(data, dict):
+            cls._validate_system_folder(data)
+            cls._validate_fiat_path(data)
+            cls._validate_sfincs_path(data)
+        return data
```
LuukBlom, line 32: Could you add some checks after initializing the settings here?

I think it would be nice to add a separate model_validator(mode="after") that actually checks that the computed fields for fiat_path and sfincs_path are pointing to the .exe/binaries that exist and raises if they dont.

LuukBlom (changes requested):
I love the new config module! 

I only have 1 real comment (about actually validating the paths point to exe/binaries that exist.

Before merging this PR, we need to remove all calls to the old config module in FloodAdapt-GUI as well. (global search for `from flood_adapt.config`)

The old way using env vars is kind of clunky anyways and this does the same, but much better.

savente93:
This requires Pixi so moving it to blocked (#485)

LuukBlom:
closing and making another [PR](https://github.com/Deltares-research/FloodAdapt/pull/719) for this

PR #531, opened 2024-08-30
524 make headers optional & allow for multiple datetime formats

LuukBlom:


flood_adapt/object_model/hazard/event/event.py
```diff
@@ -99,21 +100,69 @@ def timeseries_shape(
...
+            reader = csv.reader(f, delimiter=",")
+            num_columns = len(next(reader)) - 1  # subtract 1 for the index column
+
+        if has_header is None:
+            raise ValueError(
+                f"Could not determine if the CSV file has a header: {csvpath}."
+            )
+        header = 0 if has_header else None
+        df = pd.read_csv(csvpath, index_col=0, header=header)
+
```
savente93, line 136: Reading a CSV file doesn't actually change whether you have a header or not, so I think it's much easier to just read in the csv file with pandas as it it doesn't have a header, do your checks and set the first row as the header if you've decided there is one.

flood_adapt/object_model/hazard/event/event.py
```diff
@@ -99,21 +100,69 @@ def timeseries_shape(
...
+            "%Y%m%d %H%M%S",
+            "%Y%m%d %H%M",
+        ]
+        datetime_parsed = False
+        for fmt in SUPPORTED_DATETIME_FORMATS:
+            try:
+                df.index = pd.to_datetime(df.index, format=fmt, errors="raise")
+                datetime_parsed = True
+                break
+            except Exception:
```
savente93, line 158: No bare except please. catch the actual exceptions that `pd.to_datetime` can raise (`ValueError` and `ParseError` in this case)

tests/test_object_model/test_events.py
```diff
@@ -494,3 +494,87 @@ def test_block_discharge(test_db):
...
+                "time": time,
+                **data,
+            }
+        )
+        df.to_csv(path, index=False, header=headers, date_format=datetime_format)
+
+        df = df.set_index("time")
+        return df
+
+    def test_read_csv_ReadFileWithHeaders(self, tmp_path):
```
savente93, line 519: please use either snake_case or CamelCase but a combination of the two is a bit strange

savente93 (changes requested):
Idea looks fine, but I think there is a better way of doing this

flood_adapt/object_model/hazard/event/event.py
LuukBlom: First of all, reading a file does change whether there are headers or not. if there is no header, it uses the first value in the data column as the name of the column.
Second, regarding your suggestion, that is essentially what Im doing, I check if it has headers, and then read it accordingly? I could not find a way in pandas that returns if there is a header or not. Do you have a suggestion?

savente93: I'll make a suggestion one minute

savente93 (approved):

PR #537, opened 2024-09-03
536 time series column name format error

### LINKED ISSUE #536: time-series column name format error
### FloodAdapt version checks

### Reproducible Example

Running a scenario with downloaded water level time series fails.

### Current behaviour

The new method for reading a csv with the time-series that was introduced in https://github.com/Deltares-research/FloodAdapt/commit/84e7f0c2c05d4a9b6eed3e0f427e1deee7ec7d43 gives a column name 'data_0', 'data_1' etc. for each data column in the series. 

But the current analysis when reading in the time series dataframes, assumes the old naming, which was an integer of 1, 2, etc.
https://github.com/Deltares-research/FloodAdapt/blob/ba2ee176518294a80c1d58d04c1bffee72c378f3/flood_adapt/object_model/hazard/hazard.py#L375-L379
and
https://github.com/Deltares-research/FloodAdapt/blob/ba2ee176518294a80c1d58d04c1bffee72c378f3/flood_adapt/integrator/sfincs_adapter.py#L159-L162

### Desired behaviour

Consistensy between the new data column names and the use in the code.

### Task list

- [x] Chanes column name use in backend
- [x] Test with running scenarios

### Additional context

_No response_

panosatha:
fixes #536

### DIFF
--- a/flood_adapt/dbs_controller.py
+++ b/flood_adapt/dbs_controller.py
@@ -588,6 +588,7 @@ class Database(IDatabase):
                     )
                     return ""
                 df = input_wind_df
+                df = df.rename(columns={1: "speed", 2: "direction"}) # rename column names for consistency
             case "constant":
                 time = pd.date_range(
                     start=event.attrs.time.start_time,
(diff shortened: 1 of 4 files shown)


savente93 (approved):
LGTM

PR #541, opened 2024-09-05
Offshore model fix incl slicing error

GundulaW:
The downloaded data had longitudes around 280 degrees and the SFINCS region around -80 degrees. So I simply subtracted 360 degrees from the downloaded data, if the minimum longitude is larger than 180 degrees. Seems like a dodgy fix but it worked.


panosatha (approved):
LGTM

PR #546, opened 2024-09-17
Feat/pydantic settings for database config

LuukBlom:


flood_adapt/config.py
```diff
@@ -1,78 +1,180 @@
...
+
+    @computed_field
+    @property
+    def sfincs_path(self) -> Path:
+        return self.system_folder / "sfincs" / f"sfincs{Settings._system_extension()}"
+
+    @computed_field
+    @property
+    def fiat_path(self) -> Path:
+        return self.system_folder / "fiat" / f"fiat{Settings._system_extension()}"
```
Jaapel, line 90: Now your `sfincs_path` and your `fiat_path` always have to be the same relative to your system folder. Why not make them fully configurable?

tests/test_config.py
```diff
@@ -1,26 +1,375 @@
...
 
 
 class TestSettingsModel:
-    def test_init_no_env(self):
+    def _create_dummy_db(self, db_root: Path, name: str, system: str) -> Path:
+        ext = Settings.SYSTEM_SUFFIXES[system]
+
+        (db_root / "system" / "sfincs").mkdir(parents=True)
+        fd = os.open(db_root / "system" / "sfincs" / f"sfincs{ext}", os.O_CREAT)
+        os.close(fd)
```
Jaapel, line 20: use as context manager? `with os.open(...) as f:`?

flood_adapt/object_model/direct_impacts.py
```diff
@@ -252,9 +253,10 @@ def run_fiat(self):
         with cd(self.fiat_path):
             with open(self.fiat_path.joinpath("fiat.log"), "a") as log_handler:
                 process = subprocess.run(
-                    f'"{Settings().fiat_path}" run settings.toml',
+                    f'"{Settings().fiat_path.as_posix()}" run settings.toml',
```
Jaapel, line 254: Same goes for this path

flood_adapt/object_model/hazard/hazard.py
```diff
@@ -269,7 +269,7 @@ def run_sfincs(self):
                 sfincs_log = "sfincs.log"
                 with open(sfincs_log, "a") as log_handler:
                     return_code = subprocess.run(
-                        Settings().sfincs_path, stdout=log_handler
+                        Settings().sfincs_path.as_posix(), stdout=log_handler
```
Jaapel, line 275: and here

tests/test_config.py
```diff
@@ -75,6 +75,21 @@ def mock_system(self):
...
+        system_folder = os.environ.get("SYSTEM_FOLDER", None)
+
+        yield
+
+        if root is not None:
+            os.environ["DATABASE_ROOT"] = root
+        if name is not None:
+            os.environ["DATABASE_NAME"] = name
+        if system_folder is not None:
+            os.environ["SYSTEM_FOLDER"] = system_folder
```
Jaapel, line 103: There is a function in `utils` that I stole from SO that does this and more

flood_adapt/config.py
LuukBlom: Well, the idea of the system folder is to contain the model binaries. If we leave the model paths fully configurable, we might as well remove the system folder as a whole.

flood_adapt/config.py
```diff
@@ -167,6 +175,12 @@ def _validate_fiat_path(self):
             raise ValueError(f"FIAT binary {self.fiat_path} does not exist.")
         return self
 
+    @field_serializer(
+        "database_root", "system_folder", "sfincs_path", "fiat_path", "database_path"
+    )
+    def serialize_path(self, path: Path) -> str:
+        return path.as_posix()
```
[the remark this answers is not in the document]
LuukBlom: pydantic-settings cant apparently serialize a Path object without us manually defining a serializer like this. 
We could also just  do str(path) but that gives different results depending on the system. 
so, to be platform independent I chose to do as_posix() to have all platforms output be the same. 
Reading works as Path() can convert a posix-like str path into a Path obj and I personally dont like the double \\ as separators. 
Didnt think that much about it though, am I missing something that can result in a problem down the line?

flood_adapt/config.py
Jaapel: Why not? :)

flood_adapt/config.py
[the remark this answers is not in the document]
Jaapel: If pathlib can read a path on windows machine with `/` seperators then all good!

flood_adapt/config.py
LuukBlom: Now that I think about it, I dont see any use case for changing only one the model_paths or even the system folder in general. 

The only reason it is configurable in the first place is that it is stored on the FloodAdapt svn, and thereby the default location was in the database root, but that was a problem for distribution, as that meant users could accidentally delete/change/be confused, so it was made configurable and moved to _internal. But now there is this random dir floating around, that is critical for FloodAdapt to run and we can configure it, but why do we even want to.

I think it shouldnt even be configurable in the first place and we should just hard code it (or leave it configurable, but change the default) to expect the system folder in the source code (flood_adapt/system) since it is actually a critical part of the backend and without it, nothing works. 

This will make everyone's life easier:
- We can include the system folder when we make a release. (either already in the right place, or as a .zip attached)
- New models can still be added by just adding them to the system folder and setting up the path.
- When we create releases for windows & linux, we can include the correct system folder corresponding to the os.
- When we create exe / installers for windows & linux we do the same.

Jaapel: I would definitely package executables with the application.
If I have a working sphinx version on my workstation I can create a soft-link in this folder and make it work, so no worries there. With your current code, my initial problem of directly referencing a `sfincs.exe` is solved!

Jaapel (approved):
👍

PR #549, opened 2024-09-19
make deleting run output optional

LuukBlom:
quickfix for now, but a better, more permanent solution will be a part of the pydantic-settings PR (https://github.com/Deltares-research/FloodAdapt/pull/546) that is in review at the moment.


panosatha (approved):
LGTM!

PR #473, opened 2024-06-21
refactor(adapters): Add adapter interfaces and refactor SficnsAdapter to use them. Also refactor events to use forcings 

LuukBlom:
@GundulaW : I didnt get the refactor working yet (its getting quite big but im not sure how to nicely split it up into more PRs).
I have not removed the majority of old code, so the repo might seem a bit busy (and I didnt do imports yet so pretty much all that uses the Hazard class is broken, including all tests).

The low-level intelligence in the Forcing classes is not yet implemented, as I'd like to get your take on it before spending too much time. 

I'd love some feedback on the more domain related functionality (the SfincsAdapter is the main one, but also the Adapter interfaces, Forcings, EventModels & Events). for the SfincsAdapter, I tried to move the code from hazard.py to the adapter with as little chages as possible, but it is a lot of code so I probably introduced some bugs.

I added the Timeseries class I've told you about here as well, even though its less nice than I wanted it to be.. It can create any 1D timeseries from a SyntheticTimeseriesModel (with attrs = shapetype, shape_duration, peak_time, peak_value/cumulative), interpolate the data to a timestep, then write to a pdDataFrame while taking into account the start time & end time of an event, before adding it to a dataframe. I think it simplifies the use of Timeseries data for other parts of the code, but I also think its not perfect as it can only handle 1D Timeseries (Synthetic or csv). So wind(2D with speed and angle) is already more difficult, not to mention spatially varying gridded timeseries.

PR #558, opened 2024-10-14
553 Adding files to Database when importing extrenal files

LuukBlom:
This PR implements the solution I had in mind for the hazard refactor already and some more.
It is a lot of files changed, but its mostly in importing and replacing.

I added an `IObjectModel` & `IObject` class that every floodadapt object will inherit from containing the functions:
- `load_file`
- `load_dict`
- `save`
- `save_additional`

where the first 3 are the same for all objects and already implemented in the base class.
- `save` will call `save_additional` first (to save any extra files to the database, then update the attrs before saving attrs to a toml file.
- `save_additional` needs to be implemented for each subclass it is relevant and save the relevant files.
- `__init__` needs to be implemented for each subclass as well as explained in the `IObject` class.

I implemented the above for all classes in the backend + tests.

I also added flood_adapt\object_model\interface\path_builder.py to be able to build filepaths inside the database.
The database and dbs_classes use this module to create paths.
Meaning, other modules can also construct paths using that module, without requiring access to the whole database object.

I also moved function (get_object_ids) to a separate module since fiat_model uses it, but also some measures. Not sure about where to put it, but this is already better imo.  flood_adapt.object_model.direct_impact.measure.measure_helpers::get_object_ids

Should be merged together with: https://github.com/Deltares-research/FloodAdapt-GUI/pull/447


flood_adapt/dbs_classes/dbs_template.py
```diff
@@ -114,8 +114,14 @@ def copy(self, old_name: str, new_name: str, new_description: str):
                 continue
             shutil.copy(file, dest / file.name)
 
-    def save(self, object_model: ObjectModel, overwrite: bool = False):
-        """Save an object in the database. This only saves the toml file. If the object also contains a geojson file, this should be saved separately.
+    def save(
+        self,
+        object_model: ObjectModel,
+        overwrite: bool = False,
```
panosatha, line 120: Why is this False?

flood_adapt/object_model/utils.py
```diff
@@ -22,3 +23,37 @@ def write_finished_file(path: Path):
...
+    external_file = Path(external_file).resolve()
+    dst_dir = Path(dst_dir).resolve()
+    if not external_file.exists():
+        raise FileNotFoundError(
+            f"Could not import file {external_file} as it does not exist."
+        )
+    dst_dir.mkdir(parents=True, exist_ok=True)
+    shutil.copy2(external_file, dst_dir / external_file.name)
+
+    return dst_dir / external_file.name
```
panosatha, line 59: This method should return only the external_file.name, right? Since we want the database to be transferable. Else, this returns the absolute path to the file.

flood_adapt/dbs_classes/dbs_interface.py
```diff
@@ -60,7 +60,11 @@ def copy(self, old_name: str, new_name: str, new_description: str):
         pass
 
     @abstractmethod
-    def save(self, object_model: ObjectModel, overwrite: bool = False):
+    def save(
+        self,
+        object_model: ObjectModel,
+        overwrite: bool = False,
```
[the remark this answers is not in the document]
LuukBlom: This I did not change. 
It is to have as default behaviour to not overwrite existing objects in the db and raise an error instead. 

if overwrite=true we do overwrite it (used in edit() I believe)

flood_adapt/object_model/utils.py
LuukBlom: Correct, I didnt realize we wanted to support users sharing databases like that, but it makes sense. Will have a look tomorrow to see how to fix this nicely.

panosatha (changes requested):
Look good, but the path used in the objects needs to be relative and not absolute to avoid issues with database transfer.

flood_adapt/api/events.py
```diff
@@ -31,11 +28,11 @@ def get_event(name: str) -> IEvent:
 
 
 def get_event_mode(name: str) -> str:
-    filename = Database().events.get_database_path() / f"{name}" / f"{name}.toml"
+    filename = Database().events.input_path / f"{name}" / f"{name}.toml"
     return Event.get_mode(filename)
 
 
-def create_synthetic_event(attrs: dict[str, Any]) -> ISynthetic:
+def create_synthetic_event(attrs: dict[str, Any]) -> Synthetic:
```
panosatha, line 35: Now that this changed from ISynthetic to Synthetic I get see this message from PyLAnce:
Type "IObject[Unknown]" is not assignable to return type "Synthetic"
  "IObject[Unknown]" is not assignable to "Synthetic

flood_adapt/api/events.py
```diff
@@ -99,6 +96,10 @@ def create_historical_hurricane_event(attrs: dict[str, Any]) -> IHistoricalHurri
     return EventFactory.get_event("Historical_hurricane").load_dict(attrs)
 
 
+def save_event(event: IEvent) -> None:
+    Database().events.save(event)
+
+
 def save_event_toml(event: IEvent) -> None:
     Database().events.save(event)
```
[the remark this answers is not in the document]
LuukBlom: Should be removed I agree. I didnt want to delete api functions without checking their usage and then I forgot. Will fix this

flood_adapt/api/events.py
LuukBlom: Dont like these warnings either, added an issue for fixing them #564

flood_adapt/dbs_classes/dbs_template.py
```diff
@@ -103,8 +103,27 @@ def copy(self, old_name: str, new_name: str, new_description: str):
         # After changing the name and description, receate the model to re-trigger the validators
         copy_object.attrs = type(copy_object.attrs)(**copy_object.attrs.model_dump())
 
-        # Then save. Checking whether the name is already in use is done in the save function
-        self.save(copy_object)
+        EXCLUDE_SUFFIX = [".spw"]
```
panosatha, line 106: this should be ["*.spw"] for it to work

panosatha (commented):
editing a hurricane event does not seem to work. The object that is parsed to the save function does not have a reference to the cyc file

LuukBlom:
Incorporated in the hazard refactor: https://github.com/Deltares-research/FloodAdapt/pull/473

PR #587, opened 2024-11-22
chore(adapter): rename integrator submodule to adapter

savente93:
All classes in the integrator submodule are actually called adapters, so I think it makes more sense to rename this submodule. I guess we should also rename the adapters back to integrators if people prefer that, but that seems like more work, so I went with this approach first.


LuukBlom (approved):

savente93:
decided test failures are unrelated, we'll just merge and see later.

PR #588, opened 2024-11-22
chore(site): Rename objects for consistency

savente93:
Some of the objects in the site model had inconsistent names (e.g. `Floodmap_type` vs `FloodmapType`) and some of the things had default values that contradicted the type hints. I came across this while working on the ascii issue and figured I might as well just fix it while I'm here. I'll add the ascii validators in a separate PR

### DIFF
--- a/flood_adapt/object_model/interface/events.py
+++ b/flood_adapt/object_model/interface/events.py
@@ -137,11 +137,14 @@ class SurgeSource(str, Enum):
     shape = "shape"
 
 
+class SurgeShape(Enum):
+    gaussian = "gaussian"
+
 class SurgeModel(BaseModel):
     """BaseModel describing the expected variables and data types for harmonic tide parameters of synthetic model."""
 
     source: SurgeSource
-    shape_type: Optional[str] = "gaussian"
+    shape_type: SurgeShape = SurgeShape.gaussian
     shape_duration: Optional[float] = None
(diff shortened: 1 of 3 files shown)


flood_adapt/object_model/interface/events.py
```diff
@@ -137,11 +137,14 @@ class SurgeSource(str, Enum):
     shape = "shape"
 
 
+class SurgeShape(Enum):
+    gaussian = "gaussian"
+
```
LuukBlom, line 142: There is a ShapeType enum for this in this file

LuukBlom (changes requested):
LGTM,
could you remove the duplicate enum?
also. did you check the frontend repo as well for references to these classes?

flood_adapt/object_model/interface/events.py
savente93: sorry missed that

savente93:
I removed the duplicate, I'll check the FE in a minute, good shout

savente93:
Gui should be unaffected, I didn't find anything in there. tests seem to fail because of a pixi install issue. Could look into it if you want Luuk, but I'd say it's out of scope for this PR

LuukBlom (approved):

savente93:
test failures are unrelated to this PR so we can merge and fix those separately.

Issue #418, opened 2024-04-10
system crashes when observation station has a special character


panosatha:
### FloodAdapt version checks

- [X] I have checked that this issue has not already been reported.
- [X] I have checked that this bug exists on the latest version of FloodAdapt.

### Reproducible Example

Including ¨ (umlauts) in the name of an obsrvation station makes floodadapt break at the SFINCS postprosessing step.

### Current behaviour

Using special characters in the observation station name in the site.toml, breaks the workflow, since hydromt-sfincs cannot read the SFINCS model anymore (more specifically the issue is reading in the sfincs.obs file).

### Desired behaviour

Add validators in the pydantic model for the site object to make sure that special characters are not included in the names of the observations stations. This way even if the user has manually change something, the validator would be run during startup and give an informative message.

This could be an issue for other parameters!

### Task list

- [x] Check for which other parameters this is true
- [ ] Write site.toml validators
- [ ] Write tests

savente93:
how would this work if someone enters a name like "松阿察河" (river in china)? In general doing these kinds of conversions can be dangerous if you don't know how your input space is structured. For example, if we would just Latinise these we'd get into trouble with things like: 
 问 = wèn = to ask
 吻 = wěn = to kiss

Now we could require that input is ASCII encoded, but I think this will be quite an impediment to the adoption of FA outside of Anglicised regions.  

I'm assuming that the text in the `sfincs.obs` file is mostly either decorative (labels) or used as some kind of unique identifier. Would it be at all feasible to ask SFINCS to allow `utf-8` text in the `obs` file? I feel like that's the better solution here, as it would benifit the adoption of both sfincs and FA

savente93:
(just playing devils advocate here, hope that didn't come off too critical)

savente93:
having gone through the SFNICS docs as best I can, I found this page about the input formats: https://sfincs.readthedocs.io/en/latest/parameters.html#input-files Here any text based format is listed as `asc` which I take to mean ASCII which means that the answer to " Check for which other parameters this is true" is basically "any text that SFNICS accepts"

PR #597, opened 2024-11-26
feat(site): add ascii validators to make sure we have only ascii in labels

savente93:
Sfincs can only handle ascii, not UTF-8. We solve this potential issue by adding some pydantic validators that will error when the data the user presents us with has UTF-8 characters in it. This should then show an error and stop execution in the back-end.


savente93:
Fixes #418

savente93:
test failure are because of problem with team city, tests pass locally

flood_adapt/object_model/interface/site.py
```diff
@@ -44,7 +54,7 @@ class SfincsModel(BaseModel):
 
     csname: str
     cstype: Cstype
-    version: Optional[str] = ""
+    version: str = ""
```
panosatha, line 57: why is this non Optional anymore?

flood_adapt/object_model/interface/site.py
```diff
@@ -63,7 +73,9 @@ class WaterLevelReferenceModel(BaseModel):
 
     localdatum: VerticalReferenceModel
     msl: VerticalReferenceModel
-    other: Optional[list[VerticalReferenceModel]] = []  # only for plotting
+    other: list[VerticalReferenceModel] = Field(
```
panosatha, line 76: why is this non Optional anymore?

panosatha (approved):
LGTM. Only had some comments on some changes w.r.t. to the Optional attributes.

flood_adapt/object_model/interface/site.py
savente93: So the `Optional` is a bit of a misnomer here. Because we give it a default value, it is actually still optional. With the optional attribute it can only ever be none if the user actually specified `Cstype(version=None)` which would be very strange. So functionally it is always a string, and removing the optional attribute makes it easier on the type checker. So practically we don't change anything. Does that make sense?

flood_adapt/object_model/interface/site.py
savente93: same argument as above

PR #599, opened 2024-11-29
fix(object-model): Fix some warnings and mistakes in test suite

savente93:
I was fixing a few warnings, but removing calls to deprecated functions when I noticed some mistakes. 

```
return NotImplemented
```
Does has been deprecated since 3.9 and currently simply evaluates to true. I was this a few times and is wrong so I fixed that. it should either be `return False` or `raise NotImplementedError` if you want an error but in most cases this is not what you want. 

Further more I noticed that the `__eq__` functions copy a lot of objects only to throw them away again so we can exclude the name and description. This is really inefficient since were dealing with objects that can hold (potentially a lot of) data, so I wrote a different implementation that is significantly faster.


flood_adapt/object_model/hazard/event/event.py
LuukBlom, line 1: I dont like editing the objects themselves to do equality checking.

During the hazard refactor I also came across these things and wanted to fix them, so I came up with this
`excludes: Set[str] = { "name", "description"}`
`return self.attrs.model_dump(exclude=exclude) == other.attrs.model_dump(exclude=exclude)`

tests/conftest.py
```diff
@@ -89,7 +89,7 @@ def session_setup_teardown():
 
     if clean:
         restore_db_from_snapshot()
-    shutil.rmtree(snapshot_dir)
```
LuukBlom, line 92: I think we should raise this error. 
It is good to know that we havent closed all files properly.

LuukBlom (commented):
Thanks for doing this! I have 2 comments 
1. the ignore_errors=True in conftest
2. the way we compare object models

tests/conftest.py
savente93: sure, but then I'm gonna have to hunt down some errors, because this happens for me every local test run. I can at least take it out of the PR if you want

flood_adapt/object_model/hazard/event/event.py
LuukBlom: Also, the idea of the `attrs` attribute is to store the minimum required amount of information needed for an object model. 
So it doesnt have much data associated with it (only what is in the toml files in the database.)

The instance `obj` to which `attrs` belongs has functions to load/create/get/parse data by using the data defined in `attrs`, so Im not worried about copying just the attrs to be honest.

savente93: I don't like it in principle either, but the problem is that these are recursive models and contain objects containing data, so even with your solution we have to duplicate all that data just so we can ignore name and description which feels incredibly wasteful to me. it also shaves about 10min off my test suite run locally. (that' just a ball park idea, I didn't do very precise measuring) which I don't think we'll get with your solution.

tests/conftest.py
LuukBlom: Yeah, I have tried to reduce this stuff as much as possible, but hydromt adds a fileHandler to the logger if it doesnt have one (in setuplog) and then doesnt close it when errors are raised. I tried fixing this with the FloodAdaptLogging class, and the __del__ function for the SfincsAdapter and FiatAdapter, but it has never really been 100% when other stuff is also breaking.
Fixing this was quite time consuming so when I had a mostly working solution, I just left it

flood_adapt/object_model/hazard/event/event.py
savente93: > Also, the idea of the attrs attribute is to store the minimum required amount of information needed for an object model.
> So it doesnt have much data associated with it (only what is in the toml files in the database.)

hmmm, maybe I misread then. let me take another look at it to see where the savings come from

LuukBlom: Amazing, Im very interested to know how much this costs/saves performance wise!

flood_adapt/object_model/hazard/event/eventset.py
```diff
@@ -24,8 +24,28 @@ def load_file(filepath: Union[str, Path]):
...
+
+        ans = self.attrs == other.attrs
+
+        # restore
+        self.attrs.name = self_name
+        other.attrs.name = other_name
+        self.attrs.description = self_description
+        other.attrs.description = other_description
+
+        return ans
```
LuukBlom, line 51: I think we discussed this cleaner solution earlier:
```python
_self = self.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
_other = other.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
return _self == _other
```

flood_adapt/object_model/hazard/event/event.py
```diff
@@ -482,8 +482,28 @@ def add_wind_ts(self):
...
+
+        ans = self.attrs == other.attrs
+
+        # restore
+        self.attrs.name = self_name
+        other.attrs.name = other_name
+        self.attrs.description = self_description
+        other.attrs.description = other_description
+
+        return ans
```
LuukBlom, line 509: I think we discussed this cleaner solution earlier:
```python
_self = self.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
_other = other.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
return _self == _other
```

flood_adapt/object_model/hazard/hazard_strategy.py
```diff
@@ -15,8 +15,9 @@ def __init__(self, measures: list[HazardMeasure]) -> None:
...
-            return NotImplemented
-        names_1 = [measure.attrs.name for measure in self.measures]
-        names_2 = [measure.attrs.name for measure in other.measures]
+            return False
 
-        return set(names_1) == set(names_2)
+        names_1 = sorted([measure.attrs.name for measure in self.measures])
+        names_2 = sorted([measure.attrs.name for measure in other.measures])
+
+        return names_1 == names_2
```
LuukBlom, line 23: I think we discussed this cleaner solution earlier:
```python
_self = self.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
_other = other.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
return _self == _other
```

flood_adapt/object_model/scenario.py
```diff
@@ -105,9 +105,10 @@ def has_run_check(self):
...
 
-        test1 = self.attrs.event == other.attrs.event
-        test2 = self.attrs.projection == other.attrs.projection
-        test3 = self.attrs.strategy == other.attrs.strategy
-        return test1 & test2 & test3
+        return (
+            self.attrs.event == other.attrs.event
+            and self.attrs.projection == other.attrs.projection
+            and self.attrs.strategy == other.attrs.strategy
+        )
```
LuukBlom, line 114: I think we discussed this cleaner solution earlier:
```python
_self = self.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
_other = other.attrs.model_dump(exclude={"name", "description"}, exclude_none=True)
return _self == _other
```

LuukBlom (commented):
thanks for fixing these, and tbh my comments are a bit nitpicky since your solution also works.

In the refactor branch I also came across this issue and solved it a bit more generic like this in the ObjectModel base Class from which every class inherits:
https://github.com/Deltares-research/FloodAdapt/blob/ef5e8a20c44a6b3c5b7dcf80853cc62a06a3510c/flood_adapt/object_model/interface/object_model.py#L144

savente93:
closing this as I don't think it's worth rebasing this after teh refactor merge, I'll see if it's worth opening another with some of these changes

PR #602, opened 2024-12-04
feat(misc): Add docs and workflows for using Conventional commits

savente93:
To enable automatic change log generation and having a more uniform of recording the work done we are
introducing the use of conventional commits. There are docs on how to use this in `CONTRIBUTING.md`
and we added CI workflows to make sure we conform to the standard. PR titles and descriptions should conform
to the standard.


.commitlintrc.yml
```diff
@@ -0,0 +1,29 @@
...
+  body-empty:
+    level: error
+  scope:
+    level: error
+    optional: false
+    options:
+      - adapters
+      - object-model
+      - db-controller
+      - db-builder
```
LuukBlom, line 15: we use dbs to abbreviate in other parts of the code base so might be nice to keep that consistent

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+- ci
+- dep
+- doc
+- test
+
+and scope should be one of:
+- adapters
+- object-model
+- db-controller
+- db-builder
```
LuukBlom, line 27: same here

LuukBlom (approved):

PR #607, opened 2024-12-06
feat(adapters,object-model): Allow for use of flood-depth maps in FIAT

panosatha:
Changes that were required to allow the use of flood-depths files instead of water levels ones. This allows the use of area calculation method in FIAT with more accuracy.


panosatha:
this is now taken care in https://github.com/Deltares-research/FloodAdapt/pull/661

PR #609, opened 2024-12-09
dep(misc): pin jellyfish version so rust is not a requirement for installing

LuukBlom:
NOTE: FloodAdapt doesnt use jellyfish directly. 

jellyfish is a dependency of us->hydromt-fiat->flood_adapt that requires the rust toolchain to be installed to build a whl file for it. Normally, these whl files are created and uploaded by the publisher, but not all possible versions are available yet apparently. 
So pip tries to build it, and fails.

To not have to install the entire rust toolchain for a dependency of a dependency, I opted for just limiting the version of jellyfish where the whl is available. 

TODO: Remove the version pin when .whl files for more recent jellyfish versions are available.


savente93 (approved):
LGTM

PR #611, opened 2024-12-11
fix(dbs-builder): FIAT roads related bugs 

panosatha:
This solves issues related to Delft-FIAT roads exposure in the database builder. Now the exposure crs is used for the roads, and if roads are not available, the metrics and inforgraphics related to that are skipped.


Santonia27 (approved):
lgtm!

PR #620, opened 2025-01-06
feat(adapters): save floodmap geotiffs in sfincs projection

Santonia27:
floodmap geotiffs are saved in the sfincs projection


Santonia27:
fixed in another PR

PR #621, opened 2025-01-06
feat(adapters): add user provided netcdf forcing

LuukBlom:


flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -934,6 +939,14 @@ def _add_forcing_wind(
                 raise ValueError("No path to rainfall track file provided.")
             # data already in metric units so no conversion needed
             self._add_forcing_spw(wind.path)
+        elif isinstance(wind, WindNetCDF):
+            ds = wind.read()
+            # TODO timeframe
```
savente93, line 944: not sure what this TODO refers to, can you either resolve it, or make it more descriptive please?

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -976,13 +988,23 @@ def _add_forcing_rain(self, rainfall: IRainfall):
...
             ds = MeteoHandler().read(time_frame)
-            # data already in metric units so no conversion needed
+            # MeteoHandler always return metric so no conversion needed
+            ds["precip"] *= self._current_scenario.event.attrs.rainfall_multiplier
             self._model.setup_precip_forcing_from_grid(precip=ds, aggregate=False)
         elif isinstance(rainfall, RainfallTrack):
             if rainfall.path is None:
                 raise ValueError("No path to rainfall track file provided.")
             # data already in metric units so no conversion needed
+            # TODO rainfall multiplier
```
savente93, line 998: should this TODO be fixed before this is merged? if no, please make it more descriptive

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -976,13 +988,23 @@ def _add_forcing_rain(self, rainfall: IRainfall):
...
             self._add_forcing_spw(rainfall.path)
+        elif isinstance(rainfall, RainfallNetCDF):
+            ds = rainfall.read()
+            # TODO timeframe
```
savente93, line 1002: should this TODO be fixed before this is merged? if no, please make it more descriptive

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -976,13 +988,23 @@ def _add_forcing_rain(self, rainfall: IRainfall):
             self._model.setup_precip_forcing(timeseries=tmp_path)
         elif isinstance(rainfall, RainfallMeteo):
             ds = MeteoHandler().read(time_frame)
-            # data already in metric units so no conversion needed
+            # MeteoHandler always return metric so no conversion needed
```
savente93, line 991: Even for user provided forcings?

flood_adapt/object_model/hazard/forcing/netcdf.py
```diff
@@ -0,0 +1,28 @@
...
+@staticmethod
+def validate_netcdf_forcing(
+    ds: xr.Dataset, required_vars: set[str], required_coords: set[str]
+) -> xr.Dataset:
+    """Validate a forcing dataset by checking for required variables and coordinates."""
+    if not required_vars.issubset(ds.data_vars):
+        missing_vars = required_vars - set(ds.data_vars)
+        raise ValueError(
+            f"Missing required variables for netcdf forcing: {missing_vars}"
+        )
```
savente93, line 15: Should we also check that the required vars are not empty?

savente93: and for that matter, if we always use metric internally, do we need to check that it is in metric?

flood_adapt/object_model/hazard/forcing/rainfall.py
```diff
@@ -111,3 +113,29 @@ def save_additional(self, output_dir: Path | str | os.PathLike) -> None:
...
+    def read(self) -> xr.Dataset:
+        ds = xr.open_dataset(self.path)
+        required_vars = {"precip"}
+        required_coords = {"time", "lat", "lon"}
+        return validate_netcdf_forcing(ds, required_vars, required_coords)
+
+    def save_additional(self, output_dir: Path | str | os.PathLike) -> None:
+        if self.path:
+            output_dir = Path(output_dir)
+            if self.path == output_dir / self.path.name:
```
savente93, line 133: you might want to normalize the path here so that `/foo/bar/` and `/foo/bar/../bar` will end up as the same

flood_adapt/object_model/hazard/forcing/wind.py
```diff
@@ -138,3 +140,29 @@ class WindMeteo(IWind):
...
+    def read(self) -> xr.Dataset:
+        ds = xr.open_dataset(self.path)
+        required_vars = {"wind10_v", "wind10_u", "press_msl"}
+        required_coords = {"time", "lat", "lon"}
+        return validate_netcdf_forcing(ds, required_vars, required_coords)
+
+    def save_additional(self, output_dir: Path | str | os.PathLike) -> None:
+        if self.path:
+            output_dir = Path(output_dir)
+            if self.path == output_dir / self.path.name:
```
savente93, line 160: you might want to normalize the path here so that /foo/bar/ and /foo/bar/../bar will end up as the same

flood_adapt/object_model/hazard/forcing/netcdf.py
savente93: I think sfincs also requires a particular ordering in the spacial dimensions, so we should check that too. iirc it was `[y,x]` but that I'm not totally sure about

savente93 (commented):
Couple small comments, will be happy to approve once those are resolved

flood_adapt/adapter/sfincs_adapter.py
LuukBlom: MeteoHandler is a class part of FloodAdapt, it downloads data from external APIs using cht_meteo. It will always return metric and users cannot affect that at all

flood_adapt/adapter/sfincs_adapter.py
LuukBlom: Yeah this is one that needs to be resolved in general, I'll make an issue for it.
For every rainfall forcing, we add the forcing, and then multiply with the rainfall_multiplier users can set in the event. 

However, for forcings with source==Track, we cannot easily change the rainfall data. It is completely calculated by cht_cyclones, where it takes a cyclone track, creates a .spw file (deltares made data format of a moving 2D grid represeneting the cyclone and the wind/rainfall fields it creates). To add that to a SFINCS model, we only add the path to that file and set a config in the sfincsmodel. 

Could be the easiest to just log a warning that the multiplier doesnt do anything for forcing from Track.

savente93 (dismissed):
Looking good now :)

LuukBlom:
did not feel like dealing with the merge conflicts.

cherry picked the relevant commits to: https://github.com/Deltares-research/FloodAdapt/pull/626 and closing this

PR #623, opened 2025-01-07
feat(dbs-builder): clip exposure to hazard extent

Santonia27:
Clip the exposure to the hazard extent


flood_adapt/database_builder/create_database.py
```diff
@@ -1635,6 +1642,69 @@ def save_site_config(self):
...
+        -------
+        None
+        """
+        gdf = self.fiat_model.exposure.get_full_gdf(
+            self.fiat_model.exposure.exposure_db
+        )
+        crs = gdf.crs
+        sfincs_extend = self.sfincs.region
+        sfincs_extend = sfincs_extend.to_crs(crs)
+        sfincs_geom = sfincs_extend.unary_union
```
panosatha, line 1669: I see that the unary_union attribute is deprecated. 
You could change to sfincs_extend.union_all()

flood_adapt/database_builder/create_database.py
```diff
@@ -1635,6 +1642,69 @@ def save_site_config(self):
...
+
+        # Clip the fiat region
+        clipped_region = self.fiat_model.region.clip(sfincs_extend)
+        self.fiat_model.geoms["region"] = clipped_region
+
+        # Clip the building footprints
+        self.fiat_model.building_footprint = self.fiat_model.building_footprint[
+            self.fiat_model.building_footprint["geometry"].within(sfincs_geom)
```
panosatha, line 1677: why use the sfincs geom here if you have created the clipped_region?

flood_adapt/database_builder/create_database.py
```diff
@@ -1635,6 +1642,69 @@ def save_site_config(self):
...
+        ]
+        bf_fid = self.fiat_model.building_footprint["BF_FID"]
+        fieldname = "BF_FID"
+
+        # Clip the exposure geometries
+        # Filter buildings and roads
+        if gdf["Primary Object Type"].str.contains("road").any():
+            gdf_roads = gdf[gdf["Primary Object Type"].str.contains("road")]
+            gdf_roads = gdf_roads[gdf_roads["geometry"].within(sfincs_geom)]
```
panosatha, line 1686: do we want to keep roads that fall completely in the geom, or do we we want to cut the roads so they fall in the geom?

flood_adapt/database_builder/create_database.py
```diff
@@ -1635,6 +1642,69 @@ def save_site_config(self):
...
+            gdf_buildings = gdf[~gdf.isin(gdf_roads)]
```
panosatha, line 1687: is this the best way to get the building subset? Why not use something similar to above, like:
gdf[~gdf["Primary Object Type"].str.contains("road")]

flood_adapt/database_builder/create_database.py
Santonia27: tbh everythime I tried to use the clip function it would be the road segments that partially fall out. So I haven't found a way to clip them properly. So, the within function guarantees at least for now the the geoms are not exceeding the hazard

flood_adapt/database_builder/create_database.py
```diff
@@ -1635,6 +1642,69 @@ def save_site_config(self):
...
+            gdf_buildings = gdf_buildings[gdf_buildings[fieldname].isin(bf_fid)]
```
panosatha, line 1688: why is this step necessary? There might be cases that building objects do not have a footprint, right?

Santonia27: I moved the function below the BF creation so all buildings should have a footprint if I understand the function correctly. Also running the different tests worked out, always the same output

flood_adapt/database_builder/create_database.py
Santonia27: hm it works. Do you see any issues the function could cause this way? or is it just a different way of writing it

flood_adapt/database_builder/create_database.py
```diff
@@ -1635,6 +1642,69 @@ def save_site_config(self):
...
+            idx_roads = self.fiat_model.exposure.geom_names.index("roads")
+            self.fiat_model.exposure.exposure_geoms[idx_buildings] = gdf_buildings[
+                ["Object ID", "geometry"]
+            ]
+            self.fiat_model.exposure.exposure_geoms[idx_roads] = gdf_roads[
+                ["Object ID", "geometry"]
+            ]
+        else:
+            gdf = gdf[gdf[fieldname].isin(bf_fid)]
+            self.fiat_model.exposure.exposure_geoms[0] = gdf[["Object ID", "geometry"]]
```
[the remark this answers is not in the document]
Santonia27: but this can only be done if there are no roads as the roads do not have a bf_id. Also exposure_geoms[0] is usually only the buildings and geoms[1] the roads. So it would mix it up if it is not in the else block

flood_adapt/database_builder/create_database.py
```diff
@@ -1635,6 +1638,70 @@ def save_site_config(self):
...
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        None
+        """
+        gdf = self.fiat_model.exposure.get_full_gdf(
```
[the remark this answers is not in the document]
Santonia27: it doesn't run without the  self.fiat_model.exposure.exposure_db. 
    def get_full_gdf(
        self, df: pd.DataFrame
    ) 
the function itself needs the df as input. I havent looked into the function, if it can be written differently but also there's no budget for that

flood_adapt/database_builder/create_database.py
panosatha: The steps that makes the building footprints does not guarantee that all buildings points will have one. At least not for cases where the points are not derived from OSM.

Santonia27: I see. Then I will rewrite it.
In the future there could be a function to create a default footprint if there is non. 
Maybe a default sized square for commercial, residential and industrial

panosatha: I have added something like this as a postprocessing step to make sure that we can show something for all buildings. But to be honest from a FIAT point of view, I dont see why this matters. FIAT is ignorant about the building footprints, right? It will only use what is in the shapefiles attached to the exposure objects. In the case of FloodAdapt current version, this is always points. That is the only geometry that should affect what happens when FIAT is doing calculations. Obviously, it would be nice if the building footprints file is clipped to the hazard, but I don't think it is necessary.

flood_adapt/database_builder/create_database.py
panosatha: the .isin() methods checks if each cell in the dataframe that it is applied on, has a value that is in the dataframe given. So to be honest I am not even sure if it works correctly. Here, we want to check something per row.

Santonia27: hmm it was working fine in all the runs but i can change it

flood_adapt/database_builder/create_database.py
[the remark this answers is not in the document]
panosatha: yes I had seen then original method that why I asked. It was more of a comment that this should probably not be like this, but indeed no budget to look into this right now :)

flood_adapt/database_builder/create_database.py
panosatha: I did some checks and it seems that indeed something was going wrong with the indexing. At least for my case in Maryland, some buildings that were in the hazard regions were still clipped out.

flood_adapt/database_builder/create_database.py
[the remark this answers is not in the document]
Santonia27: so sad :')

flood_adapt/database_builder/create_database.py
Santonia27: okay, it's updated 👍

flood_adapt/database_builder/create_database.py
Santonia27: Fiat doesnt need it but I was told that FA always needs BF for the visual output right?

flood_adapt/database_builder/create_database.py
Santonia27: I wrote the code to clip the region after so I used the sfincs geoms first. But I can change it to the region

flood_adapt/database_builder/create_database.py
panosatha: Yes, but it doesn't matter if they are clipped or not. If there is and FIAT object with a BF_FID that is in the building footprints it will be used for the visualization. If not a standard shape will be assumed.

LuukBlom (approved):
LGTM

PR #624, opened 2025-01-07
feat(adapters): save floodmap geotiffs in sfincs projection

Santonia27:
reproject floodmap geotiff to sfincs projection.

I double checked that the outputs with and without reprojection are the same


panosatha (approved):
LGTM!

PR #626, opened 2025-01-08
feat(object-model): Add support for user provided netcdf forcing files

LuukBlom:
At the request of Willem, we now support (in the same way as CSV files) a way for users to create NETCDF forcings and add them to Sfincs.


savente93 (approved):
LGTM

PR #645, opened 2025-01-15
feat(dbs-builder): concat buildings and roads to one gdf

Santonia27:
before the buildings and roads where saved in geometries but the full gdf was not updated so the exposure_db was with the unlcipped data


panosatha (approved):
LGTM

PR #646, opened 2025-01-16
Fix(object-model): remove default events and forcings 

LuukBlom:
This also includes some fixes accessing the site in some places we missed in the earlier PR


savente93 (approved):
LGTM

PR #654, opened 2025-01-17
Fiat adapter refactor

panosatha:


flood_adapt/adapter/fiat_adapter.py
```diff
@@ -195,216 +729,301 @@ def apply_population_growth_new(
...
+            The name of the area. It returns the aggregation area name if the selection type is "aggregation_area",
+            the polygon file name if the selection type is "polygon", and "all" for any other selection type.
+        """
+        area = (
+            measure.attrs.aggregation_area_name
+            if measure.attrs.selection_type == "aggregation_area"
+            else measure.attrs.polygon_file
+            if measure.attrs.selection_type == "polygon"
+            else "all"
+        )
```
LuukBlom, line 819: This works, but I find code like below much more intuitive. (this is nitpicky so do with it what you want)
```python
if measure.attrs.selection_type == "aggregation_area":
    area = measure.attrs.aggregation_area_name
elif measure.attrs.selection_type == "polygon":
    area = measure.attrs.polygon_file
else:
    area = "all"
```

flood_adapt/adapter/fiat_adapter.py
LuukBlom, line 1: This looks exactly like how I had envisioned it! Very nice, I dont have any comments on structure / design!

One thing that we can do in this PR or in another PR:
I remember discussing the idea of making an enum for the fiat column names (Object ID, Primary Object Type, etc) and use the enum values as keys for the Dataframes and dictionaries. 
That way we can easily switch the Delft-FIAT/hydromt-fiat versions without having to change all occurences of them in the code.

flood_adapt/object_model/impact/measure/impact_measure.py
LuukBlom, line 1: Im not so convinced for the need of this class to be honest. (If we really want this class, make sure it inherits from IMeasure, validates it's MeasureType.is_impact(), and then let the other impact measures only inherit from this class and not the IMeasure class)

Multiple inheritance (inheriting from multiple parents at once) is not even allowed in most programming languages as it can lead to ambiguity. In python it is allowed, but if there is a simpler way of solving the problem, Im always for not using multiple inheritance.

But anyways, lets discuss!

flood_adapt/adapter/fiat_adapter.py
panosatha: Yes, this still in the list. I was thinking to directly implement that when doing the PR for the new FIAT executable, but I might as well have it ready.

flood_adapt/object_model/impact/measure/impact_measure.py
panosatha: Good point. I will update this accordingly

flood_adapt/adapter/fiat_adapter.py
panosatha: done

LuukBlom (approved):

PR #666, opened 2025-01-30
feat(adapters):update to latest delft fiat release dod

### LINKED ISSUE #651: Update to latest Delft-FIAT release (DoD)
* REplace existing references to old column names with an interface
* Intermediate interface that contains the formats the FIAT needs, and in teh future when it updates, only that needs updating.

Santonia27:


flood_adapt/adapter/fiat_adapter.py
```diff
@@ -91,6 +100,34 @@ def __init__(
         self.config_base_path = config_base_path
         self.exe_path = exe_path
         self.delete_crashed_runs = delete_crashed_runs
+        self.fiat_columns_dict = {
```
panosatha, line 103: why is this needed? You redefine all the names bellow in the fiat_output_mapping dictionary, right?

flood_adapt/adapter/fiat_adapter.py
```diff
@@ -416,10 +463,12 @@ def postprocess(self, scenario):
...
                 "templates", "infometrics", "metrics_additional_risk_configs.toml"
             )
+
             with open(config_path, mode="rb") as fp:
                 config = tomli.load(fp)["flood_exceedance"]
-            self.add_exceedance_probability(
-                column=config["column"],
+
+            metrics_exposure = self.add_exceedance_probability(
+                column=FiatColumns.inundation_depth,
```
panosatha, line 471: this is not always the inundation depth. It is provided in the config and could be the total damage for example. So probably the mapping dictionary should be used here as well?

flood_adapt/adapter/fiat_adapter.py
```diff
@@ -416,10 +463,12 @@ def postprocess(self, scenario):
             config_path = scenario.database.static_path.joinpath(
                 "templates", "infometrics", "metrics_additional_risk_configs.toml"
             )
+
             with open(config_path, mode="rb") as fp:
                 config = tomli.load(fp)["flood_exceedance"]
-            self.add_exceedance_probability(
-                column=config["column"],
+
+            metrics_exposure = self.add_exceedance_probability(
```
panosatha, line 470: why is this called metrics exposure? It is the output table per object right?

flood_adapt/adapter/fiat_adapter.py
```diff
@@ -1096,11 +1183,16 @@ def add_exceedance_probability(
...
+        fiat_results_df = self.map_risk_metrics(fiat_results_df)
+
+        return fiat_results_df
 
     def create_infometrics(
-        self, metric_config_paths: list[os.PathLike], metrics_output_path: os.PathLike
+        self,
+        metric_config_paths: list[os.PathLike],
+        metrics_output_path: os.PathLike,
+        metrics_exposure: pd.DataFrame,
```
panosatha, line 1195: doc strings need to be updated when arguments change

tests/test_adapter/test_fiat_adapter.py
panosatha, line 1: I think in the test it would be nice to use the class from the Fiat Adapter when calling column names

flood_adapt/adapter/fiat_adapter.py
Santonia27: because I only need the output mapping for the metrics after calculating the metrics.
Beforehand to actually do the calculation I used the self,fiat_columns_dict

flood_adapt/adapter/fiat_adapter.py
Santonia27: hmmm maybe mapped_exposure_df?

PR #669, opened 2025-01-31
fix(adapter): SfincsAdapter produces floodmap in correct units

LuukBlom:
Some tests broke so still fixing that


LuukBlom:
While doing stuff for this PR, I noticed that running an offshore model with a spw file as forcing input gives errors at the moment.
This was being obscured so I changed some of the exception raising we do wrt running scenarios and added a test to reproduce this error. 
Will make an issue for this

PR #678, opened 2025-02-12
feat(dbs-builder): Allow for configuring new aggregation areas

panosatha:
A aggregation areas option has been added in the database builder config when additional aggregation areas should be added.


LuukBlom (approved):
LGTM

PR #688, opened 2025-02-19
fix(adapters): updated waterlevel references in Sfincs adapter 

LuukBlom:
now all waterlevel forcings that are added to the model are updated with +msl and -localdatum

How it used to be implemented: 
https://github.com/Deltares-research/FloodAdapt/blob/3cef1953d14f2eba8cde624a05fe3bd369210b21/flood_adapt/object_model/hazard/hazard.py#L432C12-L437C14


panosatha (approved):
LGTM!

PR #689, opened 2025-02-19
fix(dbs-builder): Corrected how the existense of rivers in the SFINCS model is checked

panosatha:
Before the "dis" forcing was looked for in the HydroMT SFINCS model, but that will not work if the forcing files (.dis) are not there, and only the location of the rivers (.src) is. 

So now a check is done for the actual .src file.


LuukBlom (approved):
LGTM

PR #690, opened 2025-02-19
fix(dbs-builder): Small bug fixes in database builder

panosatha:
- small correction in aggregation area id when region is used
- ensure no issues when geometries with no exposure csv rows are present
- corrected small bug with SVI existense check
- do not save the index in the output csv of the scenario
- correct error message when scenario run fails

### DIFF
--- a/flood_adapt/database_builder/create_database.py
+++ b/flood_adapt/database_builder/create_database.py
@@ -699,10 +699,9 @@ class DatabaseBuilder:
                     )
                     footprints_found = True
             if not footprints_found:
-                self.logger.error(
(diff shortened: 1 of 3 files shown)


flood_adapt/database_builder/create_database.py
```diff
@@ -827,6 +835,7 @@ def read_fiat(self):
                 self.logger.error(
                     "No aggregation areas were available in the FIAT model and no region geometry file is available. FloodAdapt needs at least one!"
                 )
+                raise ValueError
```
LuukBlom, line 838: add log msg to the value error as well

LuukBlom (approved):
LGTM, 1 comment

PR #695, opened 2025-02-26
fix(adapters): Added option to get damage types from fiat model

panosatha:
So far the damage types were hardcoded in the adapter to "Structure" and "Content". But these can sometimes be different or have different names in the FIAT model. Now, these values are extracted from the FIAT exposure.


panosatha:
this should be merged after fiat_toolbox is updated to a new tag

Santonia27 (approved):
nice improvement

PR #699, opened 2025-03-03
dep(adapters): Update meteo handler to new version of cht_meteo

panosatha:
The latest changes in cht_meteo (v0.3.1) need some update in FA in the MeteoHandler class. Tests related to the class methods were updated as well.

### DIFF
--- a/flood_adapt/object_model/hazard/forcing/meteo_handler.py
+++ b/flood_adapt/object_model/hazard/forcing/meteo_handler.py
@@ -44,7 +44,8 @@ class MeteoHandler(IMeteoHandler):
         ds = self.dataset.collect(time_range=time_range)
 
         if ds is None:
-            self.download(time_range)
+            self.download(time)
+            ds = self.dataset.collect(time_range=time_range)
 
         if ds is None:
             raise FileNotFoundError(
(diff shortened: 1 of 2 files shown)


flood_adapt/object_model/hazard/forcing/meteo_handler.py
```diff
@@ -25,86 +20,38 @@ def __init__(self, dir: Optional[Path] = None, site: Optional[Site] = None) -> N
...
 
-        # Create an empty list to hold the datasets
-        datasets = []
-        nc_files = sorted(glob.glob(str(self.dir.joinpath("*.nc"))))
+        if ds is None:
+            self.download(time)
+            ds = self.dataset.collect(time_range=time_range)
 
-        if not nc_files:
+        if ds is None:
```
LuukBlom, line 50: Not sure about the cht_meteo implementation so this could be fixed in there, but what happens if there are downloaded files for a part of but not the complete time_range?

Will `self.dataset.collect(time_range=time_range)` return `None`? or will it return a partially filled dataset?

If so, we should add some checks here still, making sure the entire time_range has values in the dataset

LuukBlom (approved):
LGTM!

1 question on what happens when some of the time_range is already downloaded.

flood_adapt/object_model/hazard/forcing/meteo_handler.py
panosatha: that is a good one. It seems that it returns whatever it finds in the time range. I made an issue in cht_meteo. For now I changed the read method so it always download the requested times, since in cht_meteo there is check to avoid redownloading the same files.

LuukBlom (approved):
LGTM

PR #701, opened 2025-03-05
 fix(object-model): update dependencies

Santonia27:


pyproject.toml
```diff
@@ -29,14 +29,14 @@ dependencies = [
     "cht-meteo==0.3.1",
     "cht-observations",
     "cht-tide",
-    "dask<2024.7.0",  # The last version that still supports pandas 1.x, which we need until fiat-toolbox and noaa_cops become compatible with pandas 2.x
+    "dask==2024.11.2,",  # The last version that still supports pandas 1.x, which we need until fiat-toolbox and noaa_cops become compatible with pandas 2.x
```
LuukBlom, line 32: dask is not a direct dependency of FloodAdapt, but a dependency of some other dep.
I added this line because our dependencies didnt provide version limitations on what versions of pandas they require/support, which broke our env.
 
latest fiat-toolbox supports pandas 2.x now
latest noaa_coops v0.4.0 now also supports pandas 2.x

Is there a reason you pin this specific version? 
If not, I'd say remove the whole line and let the env resolver figure it out!

LuukBlom: Ok, lets remove dask from the dependencies altogether then

Santonia27: AttributeError: module 'dask.dataframe.core' has no attribute 'DataFrame'
downloading the newest Dask runs into this error unfortunately

PR #710, opened 2025-03-12
fix(object-model): fixed saving of csv/track/file-based forcings

LuukBlom:
In tomls, only file names are saved now and all files are assumed to be relative to the event toml.
When loading the events and forcings, we set up the paths correctly.


panosatha (dismissed):
Approved, but good to make an issue on how to deal with the duplication of the save_additional method for the forcing classes.

PR #715, opened 2025-03-13
fix(object-model): implement zero waterlevel reference solution

LuukBlom:
This PR implements the solution to the waterlevel referencing problem that has been discussed.
closes [#511](https://github.com/Deltares-research/FloodAdapt-GUI/issues/511)
Frontend PR: [#521](https://github.com/Deltares-research/FloodAdapt-GUI/pull/521)

I've also included - in a similar way to [the test_db input creation at startup](https://github.com/Deltares-research/FloodAdapt/blob/76ab1805425aca6496e38529e7e1c404f08b9e26/tests/data/create_test_database.py) - a function that creates the config folder used for testing.

For reviewing, I recommend first looking at the files with new code: 
- flood_adapt/object_model/interface/config/sfincs.py
- flood_adapt/object_model/interface/config/site.py
- tests/data/create_test_static.py

All other files have changes to accomodate the changes in these files.


flood_adapt/object_model/interface/config/sfincs.py
```diff
@@ -85,52 +85,64 @@ class FloodmapType(str, Enum):
...
-            self.localdatum.height -= self.msl.height
-            for other in self.other:
-                other.height -= self.msl.height
-            self.msl.height.value = 0.0
-        else:
-            self.msl.height -= self.localdatum.height
-            for other in self.other:
-                other.height -= self.localdatum.height
-            self.localdatum.height.value = 0.0
+    def main_reference_should_be_in_datums_and_eq_zero(self):
```
panosatha, line 139: Not sure if this validator is mandatory. If we assumed that the reference should be zero then we should only provide its name to use for the plots. Having it in the datums with a zero value does not provide any extra information.

flood_adapt/object_model/interface/config/sfincs.py
```diff
@@ -185,3 +211,31 @@ def read_toml(path: Path) -> "SfincsModel":
             toml_contents = load_toml(fp)
 
         return SfincsModel(**toml_contents)
+
+    @model_validator(mode="after")
+    def ensure_references_exist(self):
+        datum_names = list(self.water_level.datums.keys())
```
panosatha, line 217: datums should better be a list of DatumModels and get the names by accessing the name attribute to avoid assuming that the dict key and name attribute are the same. This would demand some extra methods and validators to make sure that names are unique and get the index based on name.

flood_adapt/object_model/hazard/forcing/plotting.py
```diff
@@ -207,25 +207,29 @@ def plot_waterlevel(
         x_title = "Time"
 
     # Plot actual thing
-    fig = px.line(data + site.attrs.sfincs.water_level.msl.height.convert(units))
+    overland_datum = site.attrs.sfincs.water_level.datums[
+        site.attrs.sfincs.config.overland_model.reference
+    ]
+    overland_datum_height = overland_datum.total_height.convert(units)
+    fig = px.line(data + overland_datum_height)
```
panosatha, line 214: this should use the tide_gauge reference and not the overland_model one! Don't forget to correct the names below in the plots.

flood_adapt/object_model/hazard/forcing/plotting.py
```diff
@@ -207,25 +207,29 @@ def plot_waterlevel(
...
-        for wl_ref in site.attrs.sfincs.water_level.other:
-            fig.add_hline(
-                y=wl_ref.height.convert(units),
-                line_dash="dash",
-                line_color="#3ec97c",
-                annotation_text=wl_ref.name,
-                annotation_position="bottom right",
-            )
+
+    for wl_ref in site.attrs.sfincs.water_level.datums.values():
```
panosatha, line 225: since here you loop through all datums, you could delete the plot above, right?

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -1331,12 +1327,18 @@ def _set_waterlevel_forcing(self, df_ts: pd.DataFrame):
                 df_ts[i + 1] = df_ts[name]
             df_ts.columns = list(range(1, len(gdf_locs) + 1))
 
-        # Add difference between msl and local datum to the water level
-        df_ts += self.settings.water_level.msl.height.convert(
-            us.UnitTypesLength(us.UnitTypesLength.meters)
-        ) - self.settings.water_level.localdatum.height.convert(
+        # Add difference between FloodAdapt datum and sfincs datum to the water level
+        main_reference_height = self.settings.water_level.datums[
```
panosatha, line 1331: this is not needed if ref is assumed to be zero

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -1331,12 +1327,18 @@ def _set_waterlevel_forcing(self, df_ts: pd.DataFrame):
...
+        ].total_height.convert(
             us.UnitTypesLength(us.UnitTypesLength.meters)
-        )
+        )  # should this always be 0 ?
+
+        sfincs_overland_reference_height = self.settings.water_level.datums[
+            self.settings.config.overland_model.reference
+        ].total_height.convert(us.UnitTypesLength(us.UnitTypesLength.meters))
+
+        df_ts += main_reference_height - sfincs_overland_reference_height
```
panosatha, line 1341: The main_reference_height should be replaced by the tide_gauge_reference. Because there might be a chance that they are different. So, let's say that ref is MLLW. But we get the time_series from the tide_gauge relative to MSL. The way it is write now will not work.

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -1362,6 +1364,9 @@ def _add_bzs_from_bca(
     ):
         # ONLY offshore models
         """Convert tidal constituents from bca file to waterlevel timeseries that can be read in by hydromt_sfincs."""
+        if self.settings.config.offshore_model is None:
+            raise ValueError("No offshore model found in site.toml.")
```
panosatha, line 1368: this should be in the sfincs.toml now, right?

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -1377,10 +1382,15 @@ def _add_bzs_from_bca(
...
+        offshore_datum = self.settings.water_level.datums[
+            self.settings.config.offshore_model.reference
+        ]
+
+        datum_height = offshore_datum.total_height.convert(us.UnitTypesLength.meters)
         for bnd_ii in range(len(sb.flow_boundary_points)):
             tide_ii = (
                 predict(sb.flow_boundary_points[bnd_ii].astro, times)
-                + event.attrs.water_level_offset.convert(us.UnitTypesLength.meters)
+                + datum_height
```
panosatha, line 1393: I think here you need to subtract the local_model_datum, since you are forcing this to that model.

panosatha (changes requested):
There are a couple of thing to be checked again before merging.

flood_adapt/adapter/sfincs_adapter.py
LuukBlom: Tide Gauges are optional though, so what would happen when there is no tide gauge?

Also `main_reference_height` will be 0 by definition so i guess this should be removed actually.

However, you make good points and made me rethink some stuff.
I implemented that the tide gauge reference is used in `_add_forcing_waterlevels()::L1083` to convert the downloaded levels to the FloodAdapt ref, then when adding it to sfincs, we add the sfincs datum.

In my mind this would be cleanest:

- The overland datum will only be used in the function `_set_waterlevel_forcing` to do the final conversion.

- All other places where we create the timeseries data should convert to the main ref by subtracting (or adding?) their own reference. see `_add_forcing_waterlevels()`

Note that WaterlevelSynthetic and WaterlevelCSV are expected to be relative to the main ref, which will be 0 so we can skip the subtraction there.

flood_adapt/object_model/hazard/forcing/plotting.py
LuukBlom: Im confused, 
1. What if there is no tide gauge?
2. The waterlevels shown in the event window should be relative to the main reference, correct?

I have added the correction when reading the tide gauge timeseries, and removed the overland datum, since we dont

flood_adapt/object_model/hazard/forcing/plotting.py
LuukBlom: The colors are different, I've added a skip now

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -1081,7 +1080,13 @@ def _add_forcing_waterlevels(self, forcing: IWaterlevel):
             conversion = us.UnitfulLength(
                 value=1.0, units=self.settings.tide_gauge.units
             ).convert(us.UnitTypesLength.meters)
-            df_ts *= conversion
+
+            datum_height = self.settings.water_level.get_datum(
+                self.settings.tide_gauge.reference
+            ).total_height.convert(us.UnitTypesLength.meters)
+
+            df_ts *= conversion + datum_height
```
panosatha, line 1088: These is a bug here! This should be done separately, since these are two different operation, while now you are multiplying with the sum.
```suggestion
            df_ts = conversion * df_ts + datum_height
```

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -578,32 +581,26 @@ def plot_wl_obs(
...
-                for wl_ref in self.settings.water_level.other:
-                    fig.add_hline(
-                        y=wl_ref.height.convert(gui_units),
-                        line_dash="dash",
-                        line_color="#3ec97c",
-                        annotation_text=wl_ref.name,
-                        annotation_position="bottom right",
-                    )
+            for wl_ref in self.settings.water_level.datums:
+                fig.add_hline(
```
panosatha, line 597: would be nice to do the same thing you do in the GUI here and plot the ref with a black line instead of green

panosatha (changes requested):
This look a lot better. Main thing is the bug with the conversions for the Gauged water levels. 

Then for the issue with the synthetic event, I think we need config attribute that provides the reference to use for the synthetic tide creation. 

Moreover, I think we need a gui config that has list of datum names that we want to include in the plots (since it can get quite crowded and especially when you have datums that are really close they plot on top of each other)

flood_adapt/adapter/sfincs_adapter.py
LuukBlom: Oh that is interesting, good to know!
Nice find!

flood_adapt/adapter/sfincs_adapter.py
```diff
@@ -1377,10 +1406,15 @@ def _add_bzs_from_bca(
...
+        offshore_datum = self.settings.water_level.get_datum(
+            self.settings.config.offshore_model.reference
+        )
+
+        datum_height = offshore_datum.total_height.convert(us.UnitTypesLength.meters)
         for bnd_ii in range(len(sb.flow_boundary_points)):
             tide_ii = (
                 predict(sb.flow_boundary_points[bnd_ii].astro, times)
-                + event.attrs.water_level_offset.convert(us.UnitTypesLength.meters)
+                - datum_height
```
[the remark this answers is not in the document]
LuukBlom: Good point! dont know how this got in here

panosatha (approved):
LGTM!

PR #716, opened 2025-03-13
fix(adapter): conditionally check for sfincs_his.nc if there are obs_points 

### LINKED ISSUE #714: Make sfincs-his.nc non-mandatory for sfincs-completed check
### Kind of request

### Enhancement Description

Currently the sfincs_completed functions checks if the sfinsc-his.nc was created. This should only be mandatory if observation points are actually in the model. 

### Task list

- [ ] make sfincs-his non-mandatory

### Use case

_No response_

### Additional Context

_No response_

LuukBlom:
closes #714

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -57,7 +57,6 @@ quartodoc:
         - name: object_model.strategy.Strategy
         - name: misc.log.FloodAdaptLogging
         - name: misc.config.Settings
-        - name: misc.config.UnitSystem
 
   package: flood_adapt
 
--- a/flood_adapt/adapter/sfincs_adapter.py
+++ b/flood_adapt/adapter/sfincs_adapter.py
@@ -5,7 +5,7 @@ import shutil
 import subprocess
 import tempfile
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import Optional, Union
 
 import geopandas as gpd
 import hydromt_sfincs.utils as utils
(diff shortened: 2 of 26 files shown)


Santonia27:
I think you need to check the postprocessing, if it needs this -his file and make this postprocessing step optional. 
Bc I tried to make that simple change and it didnt work

Santonia27:
elif isinstance(forcing, WaterlevelModel):
            from flood_adapt.adapter.sfincs_offshore import OffshoreSfincsHandler

            if self._current_scenario is None:
                raise ValueError("Scenario must be provided to run the offshore model.")

            df_ts = OffshoreSfincsHandler().get_resulting_waterlevels(
                scenario=self._current_scenario
            )
            if df_ts is None:
                raise ValueError("Failed to get waterlevel data.")

LuukBlom:
Could you explain what the problem is a bit more @Santonia27?

Notes:
- We dont create or specify that this file needs to be created. 
- hydromt-sfincs/SFINCS just creates it, and we use it.
- sfincs_his.nc is not only used for obs_points/postprocessing, it is required for getting the Offshore waterlevels to be used as input for the overland model.

LuukBlom:
Uhm, I dont understand what the problem is we are solving. 
Am I wrong in thinking that the `sfincs_his.nc` file should always be created when running sfincs?

GundulaW (approved):
Looks like what I had in mind. See my last comment in the conversation

LuukBlom:
Closing this, since its already in main somehow

Issue #734, opened 2025-04-14
Decide on API notebook storylines and assign roles


kathrynroscoe:
### FloodAdapt version checks

- [x] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/FloodAdapt)

### Kind of issue

Docs are missing

### Documentation problem

Discuss what we want to cover in the notebooks nd assign roles

kathrynroscoe:
Create environment and database so people just run these online.

LuukBlom:
write these notebooks:

- db builder - Sarah
- simple scenario - Sarah
- events - Luuk
- projection - Panos
- measures + strategies - Sarah/Luuk
- database manipulation - Luuk
- complex scenario(s)  - Sarah
- event set / risk - Gundula
- benefits - Panos

PR #746, opened 2025-05-02
doc(misc): use Minio s3 bucket for examples

LuukBlom:
## Issue addressed
Adresses #734 

## Explanation
Uploaded charleston_test to Minio
Added secret keys to gh secrets and updated docs workflow
added `docs/fetch_data.py` to retrieve database
added pixi task `fetch-data`

### DIFF
--- a/docs/fetch_data.py
+++ b/docs/fetch_data.py
@@ -1,20 +1,53 @@
 from pathlib import Path
-import boto3
-from botocore.client import Config
+from minio import Minio
+import os
 
-# The access key is created here https://s3-console.deltares.nl/access-keys
-s3 = boto3.resource(
-    's3',
-    endpoint_url='https://s3.deltares.nl',
-    aws_access_key_id='*****',
-    aws_secret_access_key='*****',
-    config=Config(signature_version='s3v4'),
-    region_name='eu-west-1'
-)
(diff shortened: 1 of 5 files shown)


docs/fetch_data.py
```diff
@@ -0,0 +1,80 @@
...
+__doc__ = """
+This script downloads the read-only public data from the MinIO bucket `flood-adapt` at `s3.deltares.nl` to the local docs/_data directory.
+
+Uploading files to the bucket can be done at this url (https://s3-console.deltares.nl), when on the Deltares VPN.
+
+To access other data in the bucket, you need to provide your own access and secret keys.
+To get non-public access keys, please contact us at floodadapt@deltares.nl
+"""
+
+READ_ONLY_ACCESS_KEY = "AZBGNdxd45VEPFp1IiGe" # read-only access key for flood-adapt/public
```
panosatha, line 16: even if this is public isn't it good practice to do some kind of encryption?

panosatha (approved):
Only a small comment on how we show the keys.

docs/fetch_data.py
LuukBlom: True, forgot to update it

panosatha (dismissed):

PR #773, opened 2025-05-21
fix(adapter): Hazard has run check was performed on SFINCS simulation and not the output

panosatha:
Fixes bug where the check for hazard run was done on the simulation and not the flooding output


LuukBlom (approved):
LGTM

PR #804, opened 2025-05-29
feat(output): DCA visualization metrics

Santonia27:
## Issue addressed
Fixes #<issue number>

## Explanation

## Checklist


panosatha:
Nice work! A couple of comments:

- I think we should not have predefined metrics type, in view of different case studies where different type of metrics might become relevant (e.g., maybe we want to show total length of roads that is flooded). Then it would be nicer if the extra metric layers and some rules are defined in the gui config.
- We can then have the logic to create and label the metrics in the backend (I already did some tests with that and it should be possible)

panosatha:
this was replaced by https://github.com/Deltares-research/FloodAdapt/pull/862

PR #832, opened 2025-06-23
refactor(dbs-builder): Allow DEBUG logging for API 

panosatha:
## Issue addressed
Fixes #<issue number>

## Explanation

## Checklist


LuukBlom (commented):
I dont think its necessary to add it to the __init__() of the db builder! 

If users want to change log level in their own script, they can just add this line to the top of their script: 

```python
FloodAdaptLogging(level=logging.DEBUG)
```

Alternatively, we could add it to the `Settings` class as well since that is already what users would put at the top of their script.

PR #844, opened 2025-06-30
feat(misc): separate generic hazard/impact configs from concrete model settings

LuukBlom:
## Issue addressed
Fixes cyclical import errors in the config module.

## Explanation
When updating documentation, I looked at the folder structure for the config module and wanted to add the import statements to the `__init__.py` similar to the other folders. However, I got cyclic import errors.

To resolve cyclical errors:
- I remembered we had this idea of separating the generic hazard/impact configs from the concrete fiat/sfincs config models. 
- The class `MeteoHandler` used to depend on the `Site`, while only needing the lat and lon from the site, so now it takes in 2 floats instead of the whole site object.

Also, I updated all import statements across FA, so thats where the 33 files changed comes from.

The files with more than just import changes are `config/__init__.py`, `config/hazard.py`, `config/impacts.py`, `config/fiat.py` and `config/sfincs.py`

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,6 @@ clean-docs = {depends-on=["clean-generated-docs", "clean-example-output", "clean
 fetch-data = { cmd = "python docs/fetch_data.py" }
 build-docs = { cmd = "quartodoc build --config docs/_quarto.yml", depends-on=[ "clean-docs" ] }
 build-interlinks = { cmd = "quartodoc interlinks", depends-on=[ "clean-docs" ], cwd = "docs" }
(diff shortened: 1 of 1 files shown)


LuukBlom:
Maybe, we should even move the concrete Model config classes to the adapter files themselves to group the config with the class itself (like db builder) and then keep only generic config settings in the `config` folder?

@panosatha what do you think? Should we move the contents of `config/fiat.py` to `fiat_adapter.py` and `config/sfincs.py` to `sfincs_adapter.py`?

panosatha (approved):
LGTM!

PR #846, opened 2025-07-03
fix(dbs-builder): Made sure crs of dem and fiat are the same

panosatha:
CRS check and reprojection in case FIAT crs is different than the SFINCS one. Additionally, delete the large subgrid folder from the sinfcs template so we dont copy it all the time.

## Issue addressed
Fixes #<issue number>

## Explanation

## Checklist
- [x] Updated documentation if needed

### DIFF
--- a/flood_adapt/database_builder/database_builder.py
+++ b/flood_adapt/database_builder/database_builder.py
@@ -1,4 +1,5 @@
 import datetime
+import gc
 import logging
 import math
 import os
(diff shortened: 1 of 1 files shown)


LuukBlom (approved):
LGTM. 

Since crs is quite important but we do not check for it explicitely and does not error when done incorrectly, should we maybe add it to the `FloodModel`or `DemModel` class? 

Perhaps, we can even add it to WaterlevelReferenceModel (and change its name to `ReferenceSystem`?)

I like being explicit better than being flexible with these kinds of things.

PR #880, opened 2025-10-08
feat(adapter): add docker for multi platform support

### LINKED ISSUE #873: Explore/document a workaround for users with Macs
### Kind of request

None

### Enhancement Description

Right now FA doesn't work on an Mac OS. Explore how Mac users could still make use of FA, maybe through a virtual environment or something like that, and then document this for Mac users. 

### Task list

see above

### Use case

_No response_

### Additional Context

_No response_

LuukBlom:
## Issue addressed
Fixes #873 

## Explanation

## Checklist


panosatha:
I did a merge from main, and added some logging statements, but I think this looks good!

THe front-end needs some changes to work though:
https://github.com/Deltares-research/FloodAdapt-GUI/pull/657

flood_adapt/adapter/docker.py
```diff
@@ -0,0 +1,310 @@
+import logging
+import re
+import subprocess
+from abc import ABC
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
```
panosatha: Dont we want to log this part in the main FloodAdapt logger so we see the info when running? 
Somehtign like this:
```suggestion
from flood_adapt.misc.log import FloodAdaptLogging

logger = FloodAdaptLogging.getLogger("Docker")
```

flood_adapt/adapter/docker.py
```diff
@@ -0,0 +1,310 @@
...
+            msg = (
+                f"Docker run failed with code {process.returncode}. "
+                f"Failed to start container `{self.name}`. Command was: {' '.join(docker_command)}. "
+                f"stdout:\n{process.stdout}. "
+                f"stderr:\n{process.stderr}. "
+            )
+            logger.error(
+                msg
+                + "Attempting to stop any existing container with the same name and retrying... "
+            )
```
panosatha: I would first log the error msg on its own and then log the warning:


```suggestion
            logger.error(msg)
            logger.warning(
                "Attempting to stop any existing container with the same name and retrying... "
            )
```

flood_adapt/adapter/docker.py
```diff
@@ -0,0 +1,310 @@
...
+                docker_command,
+                text=True,
+                capture_output=True,
+            )
+            if retry.returncode != 0:
+                msg += (
+                    f"Retry also failed with code {retry.returncode}. "
+                    f"retry stdout:\n{retry.stdout}. "
+                    f"retry stderr:\n{retry.stderr}. "
+                )
```
panosatha: why do you append to the previous msg here? This has already been logged right?


```suggestion
                msg = (
                    f"Retry also failed with code {retry.returncode}. "
                    f"retry stdout:\n{retry.stdout}. "
                    f"retry stderr:\n{retry.stderr}. "
                )
```

flood_adapt/adapter/docker.py
```diff
@@ -0,0 +1,310 @@
...
+                logger.error(msg)
+                raise RuntimeError(msg)
+
```
panosatha, line 104: I would some success log statements here:


```suggestion
            else:
                logger.info(
                    f"Retry succeeded. Container `{self.name}` started successfully after removing existing container with the same name."
                )
        else:
            logger.info(f"Container `{self.name}` started successfully.")
```

panosatha (changes requested):
I have some comments mainly related to the logging. Besides that this looks good.

flood_adapt/adapter/docker.py
LuukBlom: good catch!

Issue #635, opened 2025-01-15
Update performance requirement (DoD)


kathrynroscoe:
Make sure the simulation performance meets the required speed for interactive scenario analyses. 
Visualizations and metrics must provide useful insights for flood impact/risk analysis.
TODO: Specify real performance benchmarks based on testing with Deltares laptops (small ~1 day).

kathrynroscoe:
Use two different performance laptops for two performance metrics

kathrynroscoe:
* We need to think about the scenarios we're gonna use in this test
* Run on two computers
* Write the results/log files
* Document all specs of the models and simulation and laptop

PR #882, opened 2025-10-29
fix(dbs-controller): ensure saving geospatial files in the correct crs

LuukBlom:
## Issue addressed
Fixes [#635](https://github.com/Deltares-research/FloodAdapt-GUI/issues/635)

frontend PR: https://github.com/Deltares-research/FloodAdapt-GUI/pull/636

Most files changed are for a cleanup of the tests.
Main files to look at for review: `flood_adapt/*`

## Explanation
Any geospatial data that we save to the db is now converted to CRS 4326
I've done this by replacing the attributes refering to the geospatial data with actual `gpd.GeoDataFrame`. 

### Measure
1. In `Measure` this means that the attr `polygon_file: str | None` is replaced by `gdf: gpd.GeoDataFrame | str | Path | None`
2. Then, when the measure is read from the database, we read in the geodataframe using `Measure.read()`, which is defined in the base class `Object` to do nothing and can be extended if any files need to be read. (inverse of `save_additional`).
3. Similarly, when we save to the database, the GeoDataFrame is written to disk and the filename is stored in the toml using a `field_serializer` + the already existing `save_additional`.

### Projection
Comparing the `Projection` class with the Measure class now, we were doing almost the exact same thing with the `Projection.new_development_shapefile` attr as with the `Measure.polygon_file` attr. 
So I have just copied the exact same functionality over to the `Projection` class as well. 

### Tests:
The tests have been annoying me for a long time, they are difficult to navigate and we are testing things multiple times, which is inefficient. 
As a cleanup idea for the tests, I have moved all of the `fixtures` to `conftest.py` files. 
`pytest` automatically imports these files and exposes the fixtures to all tests at he same level or below. (similar to `__init__.py`)
While doing this, I encountered quite some bad/unused/duplicate fixtures, so I have removed those. 
The bad fixtures (ones where its just returning a dict or returning a file in the database) I have refactored, but functionally remain unchanged. 

## Checklist


flood_adapt/objects/projections/projections.py
```diff
@@ -49,25 +66,124 @@ class SocioEconomicChange(BaseModel):
     new_development_elevation : Optional[us.UnitfulLengthRefValue]
         The elevation of the new development areas. default=None.
     new_development_shapefile : Optional[str]
-        The path to the shapefile of the new development areas. default=None.
+        [DEPRECATED] The path to the shapefile of the new development areas. default=None. Use `gdf` instead.
+    gdf : Optional[gpd.GeoDataFrame | str | Path]
```
panosatha: same as my comment for the Measure

flood_adapt/dbs_classes/dbs_event.py
```diff
@@ -36,6 +36,35 @@ def get(self, name: str, load_all: bool = False) -> Event | EventSet:
         # Load event
         return EventFactory.load_file(event_path, load_all=load_all)
 
+    def copy(self, old_name: str, new_name: str, new_description: str):
```
panosatha: hmm, how come this is new? We had a copy method for events before

panosatha (commented):
All in all, I like the approach of saving everything in crs 4326 (to be honest I was under the false impression that we were already doing that :)), but I am struggling a bit with the idea of having the poylgon_file depreciated. Is it neccessary? I think we can just have a hidden attribute called _gdf that actually has the GeoDataframe of the polygon file which could be created during initialization of the object.

PR #943, opened 2026-02-06
fix(misc): Make sure env variables are updated when Setting is called

panosatha:
## Issue addressed
The recent changes included a bug that did not update the environment variables when Settings was called.

## Explanation
Made sure method is called as a validator everytime Settings is initialized.

## Checklist
- [x] Updated documentation if needed


panosatha:
this is closed since we dont want to always update the env variables when we call settings

Issue #930, opened 2026-01-30
Reduce test time


LuukBlom:
### Kind of request

Changing existing functionality

### Enhancement Description

Testing everything on teamcity takes over an hour now.
We are testing a lot of things in duplicate

I think the main areas of improvement are:
- We should use more in-memory fixtures and less file based ones. To do this, I think we need to add in-memory storage of the objects in the database. and a `read` and `write` fn on the db.
- `test_db/test_fa` fixtures: these revert the database on disk after every test. Which adds time to every test. the in-memory db will make these faster / redundant.
- We are running many scenarios in different places, and many of these scenario runs are the similar if not exactly the same. We should use fixtures to run some scenarios for output testing. 

### Task list

- [ ] add in-memory storage of the objects in the database
- [ ] add `read` and `write` to the db
- [ ] remove tests where we are testing things twice

PR #958, opened 2026-02-20
refactor(dbs-controller): database in memory

LuukBlom:
## Issue addressed

Adresses #930 

# What is in this PR

## Database changes

Instead of having the data on disk be the `truth`, now we have whatever is in-memory be the `truth`, and what is on disk is a snapshot of it. (Similar to how hydromt plugins work, you can read in template model, call some in-memory setup functions, and then call `write` to save to disk)

Main files to review to understand what this PR is about:
- `flood_adapt/dbs_classes/dbs_template.py`
- `tests/test_dbs_classes/test_dbs_template.py`

`DbsTemplate`, and thereby all of the `dbs_classes` do this now:

1. explicit `load()` function that reads in state from disk into memory
2. keeps track of mutated and deleted objects internally, and only writes changes when `flush` is called.

IO functions:
- `load`: read all data from self.input_path into memory
- `flush`: writes/deletes staged changes from memory to disk

in-memory functions:
- `get`/`copy`: no longer does any disk checks, only uses the objects that already are in-memory
- `add`: replacement of `save`, this fn performs the checks, and then stages the added object to be written in the next call to `flush`
- `delete`: no longer does the deletion from disk, this does remove it from the in-memory objects, and stages the removed obj to be deleted from disk in next call to `flush`

This means much less time spent doing IO, looping over files, running validators, reading contents/metadata from files.

## Object changes

- `save_additional`: calling this fn no longer changes paths stored in the object itself, it only ensures the additional files are written.
- `_post_load`: now responsible for updating the path variables inside of objects. This fn gets called in `flush()` when an obj is saved / mutated / overwritten.

## Testing

I have also reviewed the tests and have tried to reduce testing time, see #930 

What I have done:
- `removed duplicate testing`: there were quite some tests that were testing something not directly related to the code under test, or that were testing something that is already tested in another file. (mainly `test_flood_adapt.py` tests were just doing one liner call-forwarding to already tested code.
- `cleaned up tests`: added some helper test functions / kept test idea, but improved the approach/speed.
- `added pytest marker`: 'integration', I added this to all tests that require a full scenario run. I added this so we can easily skip slow tests when developing locally. Teamcity will always run all tests (688 in total at the moment).

Added pixi tasks:
- `pixi run tests-unit`: runs all tests that dont have the integration marker, now takes 6:30 mins locally to run 653 out of 688 tests, with ~70% coverage)
- `pixi run tests-integration`: runs all tests that have the integration marker. (the other 25 out of 688)
- `pixi run tests` still runs all tests. (unchanged)

### DIFF
--- a/flood_adapt/adapter/sfincs_adapter.py
+++ b/flood_adapt/adapter/sfincs_adapter.py
@@ -1,3 +1,4 @@
+import gc
 import logging
 import math
 import os
@@ -1739,7 +1740,9 @@ class SfincsAdapter(IHazardAdapter):
     def _get_zsmax(self):
         """Read zsmax file and return absolute maximum water level over entire simulation."""
         self._model.read_results()
-        zsmax = self._model.results["zsmax"].max(dim="timemax")
+        zsmax = self._load_and_copy_results_dataset(
+            self._model.results["zsmax"].max(dim="timemax"), "zsmax"
+        )
 
         # Convert from meters to floodmap units
         floodmap_conversion = us.UnitfulLength(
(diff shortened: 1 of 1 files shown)


panosatha (approved):
LGTM!

PR #970, opened 2026-02-27
feat(api): Added method to get existing structures from SFINCS model

### LINKED ISSUE #937: View baseline model infrastructure
### Kind of request

None

### Enhancement Description

Right now users cannot see which existing infrastructure was included in the baseline SFINCS model in their system. This enhancement will allow users to view the existing infrastructure.

### Task list

Create a 'show existing infrastructure' option, possibly on the site tab, maybe in the measures tab. 

### Use case

_No response_

### Additional Context

_No response_

panosatha:
## Issue addressed
Fixes #937 

## Explanation
Added a method to get the existing weirs from SFINCS as a geodataframe through the SFINCS adapter.

## Checklist

### DIFF
--- a/flood_adapt/dbs_classes/dbs_static.py
+++ b/flood_adapt/dbs_classes/dbs_static.py
@@ -117,6 +117,12 @@ class DbsStatic(IDbsStatic):
         return grid
 
     @cache_method_wrapper
+    def get_existing_structures(self) -> gpd.GeoDataFrame:
+        """Get existing structure geometries from model."""
+        structures = self.get_overland_sfincs_model().get_existing_structures()
+        return structures
+
+    @cache_method_wrapper
     def get_obs_points(self) -> Optional[gpd.GeoDataFrame]:
         """Get the observation points from the flood hazard model."""
         if self._database.site.sfincs.obs_point is None:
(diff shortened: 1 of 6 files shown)


LuukBlom (commented):
Nice, I think we should make the `get_existing_structures` into explicit properties that can return None as well if they are not defined in the model.

flood_adapt/dbs_classes/dbs_static.py
```diff
@@ -122,6 +122,14 @@ def get_model_grid(self):
         grid = self.get_overland_sfincs_model().get_model_grid()
         return grid
 
+    @cache_method_wrapper
+    def get_existing_structures(self) -> dict[str, Optional[gpd.GeoDataFrame]]:
+        """Get existing structure geometries from model."""
+        structures = {}
+        structures["weirs"] = self.get_overland_sfincs_model().weirs
+        structures["drainage"] = self.get_overland_sfincs_model().drainage
+        return structures
```
LuukBlom: `self.get_overland_sfincs_model()` is not cached so running it twice actually reads the template model twice.

```python
    @cache_method_wrapper
    def get_existing_structures(self) -> dict[str, gpd.GeoDataFrame | None]:
        """Get existing structure geometries from model."""
        overland = self.get_overland_sfincs_model()
        return {
            "weirs": overland.weirs,
            "drainage": overland.drainage
        } # perhaps even return copies of the structures to prevent errors/bugs when calling this and mutating the return value?
```

LuukBlom (changes requested):
small comment about reading the template model twice.
Otherwise LGTM!

PR #988, opened 2026-06-10
fix(misc): Replace DEM tile pyramids with single GeoTIFFs + harden infographic errors

panosatha:
Summary

  Backend support for the new MapLibre/Guitares GUI, which renders raster layers (topography and flood maps) as server-side image overlays via cht_tiling
  instead of client-side PNG tile pyramids. The database builder now produces a single index GeoTIFF, existing databases are migrated automatically on
  open, and the risk-infographic path now raises a clear, actionable error instead of a cryptic KeyError.

  Changes

  1. Database builder: index GeoTIFF instead of PNG tiles (database_builder.py)
  - create_dem_model no longer moves/builds static/dem/tiles PNG pyramids. It now generates a single index.tif (via hydromt_sfincs … make_index_cog) that
  maps each high-res DEM pixel to its SFINCS grid cell, alongside the subgrid DEM GeoTIFF.
  - The GUI reads DEM.tif + index.tif directly, so the deprecated tiling step (and the os/setup_tiles code path) is removed.

  2. Automatic migration of older databases (dbs_classes/database.py)
  - New _migrate_deprecated_tiles() runs once on database open and:
    a. Generates index.tif from the overland SFINCS model if it's missing (_generate_index_geotiff).
    b. Deletes the deprecated static/dem/tiles folder if present.
  - Both actions are logged as warnings (this mutates the database on disk), and all failures are caught and logged so opening a database never fails
  because of the migration — at worst the flood map doesn't render until the index is available.
  - get_topobathy_path() / get_index_path() now point at the single GeoTIFFs (static/dem/<dem>.tif, static/dem/index.tif) instead of the tile folders.
  - zsmax is now flattened in Fortran order to match the index GeoTIFF's cell numbering expected by cht_tiling's FloodMap.

  3. Clearer risk-infographic errors (adapter/fiat_adapter.py)
  - Wrapped infographic creation so a missing query field in config_risk_charts.toml (introduced by newer fiat_toolbox, which switched from hardcoded
  metric names to config-driven ones) raises an explanatory error showing exactly which lines to add, rather than a bare KeyError: 'query'.
  - A query pointing at a non-existent metric now reports which metric is missing and where to look.

### DIFF
--- a/flood_adapt/dbs_classes/database.py
+++ b/flood_adapt/dbs_classes/database.py
@@ -332,10 +332,10 @@ class Database(IDatabase):
             try:
                 self._generate_index_geotiff(index_path)
             except Exception as e:
-                logger.warning(
+                raise DatabaseError(
                     f"Could not generate the index GeoTIFF at {index_path.as_posix()}: "
                     f"{e}. The flood map will not render until it is available."
-                )
+                ) from e
 
         # 2. Remove the deprecated tile pyramids.
         if tiles_dir.exists():
@@ -362,7 +362,9 @@ class Database(IDatabase):
         from hydromt_sfincs import SfincsModel as HydromtSfincsModel
         from hydromt_sfincs.workflows.downscaling import make_index_cog
 
-        overland_root = self.static_path / "templates" / "overland"
+        overland_root = (
+            self.static_path / "templates" / self.site.sfincs.config.overland_model.name
+        )
         logger.info(
             f"Generating index GeoTIFF at {index_path.as_posix()} from SFINCS model "
             f"{overland_root.as_posix()} ..."
--- a/flood_adapt/dbs_classes/database.py
+++ b/flood_adapt/dbs_classes/database.py
@@ -410,7 +410,8 @@ class Database(IDatabase):
         Returns
         -------
         np.array
-            2D map of maximum water levels
+            1D per-cell array of maximum water levels,
+            matching the cell numbering of the index GeoTIFF
         """
         # If single event read with hydromt-sfincs
         if not return_period:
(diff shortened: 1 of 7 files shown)


flood_adapt/dbs_classes/database.py
```diff
@@ -276,26 +280,101 @@ def get_flooding_path(self, scenario_name: str) -> Path:
...
+                f"GeoTIFFs ({self.site.sfincs.dem.filename}.tif / index.tif) instead, so these tiles are no "
+                "longer used and are being deleted to avoid confusion and save space."
+            )
+            try:
+                shutil.rmtree(tiles_dir)
+            except Exception as e:
+                logger.warning(
+                    f"Could not delete the deprecated tiles folder {tiles_dir.as_posix()}: "
+                    f"{e}. It is unused and can be removed manually."
+                )
```
LuukBlom, line 353: I think we should actually raise an error when the index cant be created.
This postpones/hides the error, warnings at startup tend to be missed/ignored, especially when guitares pushed 100 messages to the terminal.

flood_adapt/dbs_classes/database.py
```diff
@@ -276,26 +280,101 @@ def get_flooding_path(self, scenario_name: str) -> Path:
...
+    def _generate_index_geotiff(self, index_path: Path) -> None:
+        """Generate the index GeoTIFF from the overland SFINCS model and the subgrid DEM.
+
+        Mirrors flood_adapt's ``database_builder.create_dem_model``: each high-resolution
+        DEM pixel is mapped to its SFINCS grid cell index. Heavy imports are done lazily.
+        """
+        from hydromt_sfincs import SfincsModel as HydromtSfincsModel
+        from hydromt_sfincs.workflows.downscaling import make_index_cog
+
+        overland_root = self.static_path / "templates" / "overland"
```
LuukBlom:
```python
overland_root = (
            self._database.static_path
            / "templates"
            / self.site.sfincs.config.overland_model.name
        )
```

flood_adapt/dbs_classes/database.py
LuukBlom: Im okay with warning if the tiles cannot be deleted

LuukBlom (dismissed):
LGTM, only minor comments

LuukBlom (dismissed):

PR #995, opened 2026-06-24
fix teamcity and fetch data

LuukBlom:
test for teamcity


docs/fetch_data.py
```diff
@@ -21,32 +15,43 @@
...
+
+
 def download_directory(client: Minio, path_in_bucket: str, output_path: Path, overwrite: bool = False, bucket_name = "flood-adapt") -> None:
     """Download a directory from a MinIO bucket to a local directory."""
     if output_path.exists() and not overwrite:
         raise FileExistsError(
             f"Output path {output_path} already exists. Use `overwrite=True` to overwrite."
         )
-
-    if not client.bucket_exists(bucket_name):
```
panosatha, line 31: Wasn't this error more specific and nice to have? Why take it out?

docs/fetch_data.py
```diff
@@ -63,33 +68,66 @@ def prepare_client(access_key: str, secret_key: str) -> Minio:
...
-        path_in_bucket="system", # requires the non-public access keys
-        output_path=data_dir / "system",
-        overwrite=True
-    )
+    for path_in_bucket in args.path_in_bucket:
+        download_directory(
+            client=client,
+            path_in_bucket=path_in_bucket.strip(),
+            output_path=args.data_dir / path_in_bucket.strip(),
+            overwrite=args.overwrite
```
panosatha, line 131: Before this was defaulting to True but now since it is not provided in the docs.yml, now it will be false. Do we want that?

LuukBlom: Should be fine, and I think False is the correct default.
In the docs.yml / gh actions, there will never be data cached, since its a fresh runner so that is fine
Its now when you try to overwrite your local cache, it will let you know by erroring.

docs/fetch_data.py
LuukBlom: I was getting this error in some way, not sure how tbh. But the bucket obviously exists, so there is something wrong here, but I didnt feel like spending more time tbh.
If the bucket doesnt exist, we will hit the ValueError below: `No objects found in ...`

